PR #4, opened 2022-11-29
KubeHunter test: add copy/paste text, bash command, esc modal, verify data

charisprose:


jacobbeasley:
@charisprose  Lets not commit chromedriver. Maybe we should add it to the .gitignore, too.

jacobbeasley:
Chrome driver is large and I don't want to make everybody download executable files (which could contain viruses or other things) when downloading our code repository.

charisprose:
Make sense. I'll add the chromedriver to the .gitignore file.

Charis Prose
Software Engineer of Intelletive
***@***.***

PR #22, opened 2022-12-16
Falco event details2

charisprose:


dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -113,12 +120,58 @@ export class FalcoDao {
...
+        const signatureLogs= knex.select()
+            .from('project_falco_logs')
+            .where('anomaly_signature', signature)
+            .andWhere('cluster_id', clusterId)
+            .andWhere('calendar_date', '>=', startDate)
+            .andWhere('calendar_date', '<=', endDate);
+
+
+        const signatureCountByDate = await knex
+            .select( [knex.raw( 'calendar_date as date, count (calendar_date)')])
```
jacobbeasley, line 158: One thing to consider is if it occurs on Monday and Wednesday, we may still want the API to return 0 for Tuesday. So, we may want to populate days with 0 results, too, in the graph (either on frontend or backend).

dash/frontend/src/app/core/services/falco.service.ts
```diff
@@ -17,77 +31,77 @@ export class FalcoService {
...
-              limit?: number,
-              page?: number,
-              selectedPriorityLevels?: string [],
-              selectedOrderBy?: string,
-              startDate?: string,
-              endDate?: string,
-              namespace?: string,
-              pod?: string,
-              image?: string
+  buildParams(clusterId: number, options?: FalcoLogOptions
```
jacobbeasley, line 57: I like how you cleaned this up. Very extensible.

dash/frontend/src/app/core/services/falco.service.ts
```diff
@@ -17,77 +31,77 @@ export class FalcoService {
...
   ): HttpParams {
     let params: HttpParams = new HttpParams();
     params = params.set('cluster-id', clusterId.toString());
 
-    if (limit){
-      params = params.set('limit', String(limit));
+    if (options?.limit){
+      params = params.set('limit', String(options?.limit));
```
jacobbeasley, line 63: Lets make sure the limit and page count are numeric.

jacobbeasley: I think you can multiply it by 1 to force it into being a number. Kyle may have another way, too. In SQL normally the limit/page count/sort by are not escaped.

dash/frontend/src/app/core/services/falco.service.ts
```diff
@@ -17,77 +31,77 @@ export class FalcoService {
...
+    if (options?.selectedPriorityLevels) {
+      options?.selectedPriorityLevels.forEach((priority: string) => {
         params = params.append('priority[]', priority);
       });
     }
 
-    if (selectedOrderBy) {
-      params = params.set('orderBy', selectedOrderBy.toString());
+    if (options?.selectedOrderBy) {
+      params = params.set('orderBy', options?.selectedOrderBy.toString());
```
jacobbeasley, line 76: Lets have a list of "allowed" order by options and not allow them to select anything else. In SQL, the order by is a field name so it isn't escaped the same way. Maybe the framework handles this for us, or maybe not, but I find that making sure its a list of allowed ordered by options is the most foolproof way to prevent sql injection.

charisprose: I have the list of order by options here https://github.com/m9sweeper/m9sweeper/blob/main/dash/frontend/src/app/modules/private/pages/falco/falco-events-list/falco-events-list.component.ts. Is it what you looking for?

jacobbeasley: No, we need to also validate it in the backend because a nefarious user could pass anything to the backend (regardless of what list we show in the frontend).

dash/backend/src/modules/falco/dao/falco.dao.ts
charisprose: I added this with JS instead of using query.

PR #24, opened 2023-01-11
Intrusion detection wip

charisprose:


dash/backend/src/email-templates/falco-log-email.hbs
```diff
@@ -0,0 +1,37 @@
...
+    td{
+        text-align: left;
+        width: 80%;
+    }
+    th {
+        text-align: left;
+        width: 20%;
+    }
+</style>
+
```
jacobbeasley, line 11: Checkout how some other apps format their emails - https://raygun.com/blog/new-error-notification-emails-packed-with-data/

Lets make a few changes here to make the email a bit nicer looking: 

1) Enclose it in a grey box 

2) At an intro before displaying message details

Hello there,

A new {Rule} error occurred in {namespace}. 

Put the contents of the falco message on ints own line. 


        <tr>
            <th colspan=2>Message</th>
        </tr>
        <tr>
            <td colspan=2>{{falcoLog.message}}</td>
        </tr>

dash/backend/src/email-templates/falco-log-email.hbs
```diff
@@ -0,0 +1,37 @@
...
+            <th>Severity</th> <td>{{falcoLog.level}}</td>
+        </tr>
+        <tr>
+            <th>Rule</th> <td>{{falcoLog.rule}}</td>
+        </tr>
+        <tr>
+            <th>Tags</th> <td>{{falcoLogTags}}</td>
+        </tr>
+        <tr>
+            <th>Message</th> <td>{{falcoLog.message}}</td>
```
jacobbeasley, line 32: May want to put message in <blockquote> tag

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +93,51 @@ export class FalcoController {
...
-            const currentUserAuth = currentUserAuthObj[0].authorities;
-            // get Falco authority from Authorityid
-            let authorityArr: AuthorityId [] = [AuthorityId.FALCO];
-            // is the user a Falco user
-            let isFalcoUser = this.authService.checkAuthority(currentUserAuth, authorityArr);
-            if (isFalcoUser) {
-                return this.falcoService.createFalcoLog(clusterId, falcoLog);
-            }
+        if (!currentUserAuthObj) {
+            return; // @TODO: Should throw access denied exception
```
jacobbeasley, line 104: Should throw UnauthorizedException if access credentials are invalid

See https://github.com/m9sweeper/m9sweeper/blob/main/dash/backend/src/modules/auth/services/ldap/generic-ldap.service.ts#L92 as an example.

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +93,51 @@ export class FalcoController {
...
         }
+
+        // get all authorities from the current user
+        const currentUserAuth = currentUserAuthObj[0].authorities;
+        let isFalcoUser = this.authService.checkAuthority(currentUserAuth, [AuthorityId.FALCO]);
+        if (!isFalcoUser) {
+            return; // @TODO: Should throw access denied exception
```
jacobbeasley, line 111: Should throw ForbiddenException if you are not allowed to do this action.

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +93,51 @@ export class FalcoController {
...
+                    return; // don't save this event - it is being filtered out
+                }
+            }
+        }*/
+
+        // Saved new falco log
+        const newFalcoLog = await this.falcoService.createFalcoLog(clusterId, falcoLog);
+
+        // look up falco settings in order to decide if we need to send email(s) to administrators
+        await this.falcoService.sendFalcoEmailAlert(clusterId, newFalcoLog);
```
jacobbeasley, line 128: This line can be removed. Instead, inside it should be called from inside of falcoService.createFalcoLog.

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +93,51 @@ export class FalcoController {
...
+    }
+
+
+    @Post(':clusterid/settings')
+    async createFalcoSetting(
+        @Param('clusterid') clusterId: number,
+        @Body() falcoSetting: FalcoSettingDto
+    ): Promise <any> {
+        console.log('backend falco controller');
```
jacobbeasley, line 137: need to remove the console.logs

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +93,51 @@ export class FalcoController {
...
+        console.log('falcosetting:', falcoSetting);
+        const result = this.falcoService.createFalcoSetting(clusterId, falcoSetting);
```
jacobbeasley, line 139: After this line it should call this.sendFalcoEmailAlert() so that whenever a new falco event is created (from anywhere in the application) it will always send an email if it is appropriate to do so.

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -192,11 +195,94 @@ export class FalcoDao {
...
+                .then(res => res?.rows[0]?.exists);
+
+            if (!subquery){
+                const result = knex
+                    .insert(instanceToPlain(falcoSetting))
+                    .into('falco_settings')
+                    .returning(['*']);
+
+                return result;
+            }else {
```
jacobbeasley, line 218: } else {

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -192,11 +195,94 @@ export class FalcoDao {
...
+                const result = knex('falco_settings').update(instanceToPlain(falcoSetting)).where('cluster_id', clusterId);
+                return result;
+            }
+        }
+
+    async findFalcoSettingByClusterId(clusterId: number): Promise<FalcoSettingDto> {
+        const knex = await this.databaseService.getConnection();
+
+        // see if any settings matches the clusterid
+        const query = await knex.select().from('falco_settings').where('cluster_id', clusterId);
```
jacobbeasley, line 228: Lets validate that this actually is returning a FalcoSettingDto. I think it may be returning an object with similar fields, but with underscores instead of camel case.

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -192,11 +195,94 @@ export class FalcoDao {
...
+            .leftJoin('user_authorities as ua', function(){
+                this.on('u.id','=','ua.user_id')
+            })
+            .whereIn('ua.authority_id', [1,2])
+        return knexnest(query)
+            .then(data => {
+                return data;
+            });
+    }
+    async addFalcoEmail(emailSentTime: number, clusterId: number, falcoSignature: string): Promise<any> {
```
jacobbeasley, line 255: should have a space above a function

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+        }
+
+        // get all admin email addresses
+        const allAdminEmail = await this.getAllAdminsToMail();
+        // To test email: const allAdminEmailArray = ['some_email_address'] and comment out forEach();
+        const allAdminEmailArray = [''];
+        allAdminEmail.forEach( element => allAdminEmailArray.push(element.email));
+
+        // Parse data from new falco log fields
+        const falcoId = newFalcoLog[0].id;
```
jacobbeasley, line 186: Why is this using array syntax? On line 172 it is defined as a FalcoDto object, not an array of FalcoDto objects.

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+        const falcoSignature = newFalcoLog[0].anomaly_signature;
+        const falcoNamespace = newFalcoLog[0].namespace;
+        const falcoSeverity = newFalcoLog[0].level;
+
+        // Parse data from falco settings json fields
+        const severityLevel = falcoSetting["severity_level"];
```
jacobbeasley, line 192: How does this work if the severity_level variable does not exist on the FalcoSettingDto class?

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+
+        if (!severityLevel.includes(falcoSeverity)) {
+            return; // no need to send email - does not match the severity level we send emails for
+        }
+
+        // Have we already sent an email ?
+        const lastEmailSentTime = await this.falcoEmailAlreadySent(clusterId, falcoSignature);
+
+        // if no email record then send an email now
+        if (lastEmailSentTime === null || lastEmailSentTime <= (Date.now() - lastEmailSentTime)) {
```
jacobbeasley, line 211: This logic does not make sense: 

lastEmailSentTime <= (Date.now() - lastEmailSentTime)

I think it ought to be 

        lastEmailSentTime <= (Date.now() - anomalyFrequency)

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+
+        // Parse data from new falco log fields
+        const falcoId = newFalcoLog[0].id;
+        const falcoSignature = newFalcoLog[0].anomaly_signature;
+        const falcoNamespace = newFalcoLog[0].namespace;
+        const falcoSeverity = newFalcoLog[0].level;
+
+        // Parse data from falco settings json fields
+        const severityLevel = falcoSetting["severity_level"];
+        const anomalyFrequency = (falcoSetting['anomaly_notification_frequency'] * 24 * 60 * 60 * 1000);
```
jacobbeasley, line 193: anomolyFrequency is a vague variable name. Lets change this to anomolyAlertFrequencyAsMilliseconds

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+        const weekDay = falcoSetting['weekday'];
+        const emailList = falcoSetting['email_list'].split(',');
+        const cleanEmailList = [];
+        for (const ele of emailList) {
+            cleanEmailList.push(ele.trim());
```
jacobbeasley, line 198: If it is empty, it should not push to the list. 

I would refactor to... 

        const cleanEmailList = emailList.map(e => e.trim()).filter(e => e !== '');

dash/backend/src/modules/falco/dao/falco.dao.ts
jacobbeasley: I believe you can use the plainToInstance function to fix this bug. 

https://github.com/m9sweeper/m9sweeper/blob/main/dash/backend/src/modules/cluster/dao/cluster.dao.ts#L38

        const query = plainToInstance(FalcoSettingDto, await knex.select().from('falco_settings').where('cluster_id', clusterId))

Or something like that. I see we are doing that elsewhere in the application.

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+                falcoNamespace,
+                falcoSignature,
+                newFalcoLog
+            );
+
+            // if email is sent then add a falco email record
+            if (emailSentTime !== null && emailSentTime !== undefined) {
+                await this.addFalcoEmail(emailSentTime, clusterId, falcoSignature);
+            } else {
+                console.log("Failed to send email!")
```
jacobbeasley, line 245: I believe instead of console.log we have a logger class that should be used. 

https://github.com/m9sweeper/m9sweeper/blob/6a09d069725d3381c79b01d3f3d14b1073b152c7/dash/backend/src/modules/command-line/services/kubernetes-cluster.service.ts#L69

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+            await this.sendFalcoEmailAndAddFalcoEmailRecord(whoToNotify, allAdminEmailArray, emailList, clusterId, falcoId, falcoSeverity, falcoNamespace, falcoSignature, newFalcoLog);
+        } else {
+            // when was the last email sent?
+            const timeDifference = Date.now() - lastEmailSentTime;
+
+            // if the last sent email was older than anomalyFrequency then send
+            if (timeDifference > anomalyFrequency) {
+                await this.sendFalcoEmailAndAddFalcoEmailRecord(whoToNotify, allAdminEmailArray, emailList, clusterId, falcoId, falcoSeverity, falcoNamespace, falcoSignature, newFalcoLog);
+            } else {
+                console.log('Email notification was already sent for this anomaly!');
```
jacobbeasley, line 221: Should use loggerService like in https://github.com/m9sweeper/m9sweeper/blob/6a09d069725d3381c79b01d3f3d14b1073b152c7/dash/backend/src/modules/command-line/services/kubernetes-cluster.service.ts#L69

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +125,126 @@ export class FalcoService {
...
+
+        // Have we already sent an email ?
+        const lastEmailSentTime = await this.falcoEmailAlreadySent(clusterId, falcoSignature);
+
+        // if no email record then send an email now
+        if (lastEmailSentTime === null || lastEmailSentTime <= (Date.now() - lastEmailSentTime)) {
+            await this.sendFalcoEmailAndAddFalcoEmailRecord(whoToNotify, allAdminEmailArray, emailList, clusterId, falcoId, falcoSeverity, falcoNamespace, falcoSignature, newFalcoLog);
+        } else {
+            // when was the last email sent?
+            const timeDifference = Date.now() - lastEmailSentTime;
```
jacobbeasley, line 215: I don't believe you need to check the timeDifference here because you are already checking it above.

dash/backend/src/modules/shared/services/email.service.ts
```diff
@@ -23,13 +24,17 @@ export class EmailService {
 
       if (this.configService.get('email.smtp.host')) {
         await this.mailerService.sendMail(mailData); // send async
```
jacobbeasley, line 26: Does this service return true/false based on whether it was successful sending the email?

dash/backend/src/modules/shared/services/email.service.ts
```diff
@@ -12,6 +12,7 @@ export class EmailService {
   ) {}
 
   async send(mailData: ISendMailOptions): Promise<any> {
```
jacobbeasley, line 14: instead of any perhaps we could specify what the response is now

dash/backend/src/modules/shared/services/email.service.ts
```diff
@@ -23,13 +24,17 @@ export class EmailService {
 
       if (this.configService.get('email.smtp.host')) {
```
jacobbeasley, line 25: If this is false, will it not send?

dash/frontend/src/app/core/services/falco.service.ts
```diff
@@ -54,6 +56,11 @@ export class FalcoService {
     return this.httpClient.get('/api/falco/download', {params});
   }
 
+  addFalcoSetting(clusterId: number, settingPayload: IFalcoSettingPayload): Observable<IServerResponse<IFalcoSettingPayload>>{
+    console.log('falco service');
```
jacobbeasley, line 60: Should use loggerService

dash/frontend/src/app/modules/private/pages/falco/falco-settings/falco-settings.component.html
```diff
@@ -31,38 +31,45 @@ <h3>Notify about Anomalies</h3>
...
-                      </li>
-                    </ul>
                 </div>
               </div>
 
               <div class="section-sublevel-1">
                 <div class="row">
                   <div>
-                    <mat-checkbox matInput formControlName="sendEmail"> </mat-checkbox>
+                    <mat-checkbox (change)="enableSummarySubLevelcheckbox($event)" matInput formControlName="sendNotificationSummary" disabled> </mat-checkbox>
```
jacobbeasley: We don't want to show Falco settings for parts we haven't implemented yet. I know we discussed commenting this out but I don't see it commented out.

dash/frontend/src/app/modules/private/pages/falco/falco-settings/falco-settings.component.html
```diff
@@ -90,7 +100,7 @@ <h3>Who to Notify</h3>
     </div>
 
     <div class="event-table col-sm-6">
-      <mat-card style="height:780px; min-width: 400px; display: grid;!important;">
+      <mat-card style="height:800px; min-width: 400px; display: grid;!important;">
```
jacobbeasley, line 104: Because this feature isn't implemented, I think we should just set "display: none" to the entire mat-card.

dash/backend/src/modules/falco/service/falco.service.ts
charisprose: Will use plainToInstance on the result of createFalcoLog to return FalcoDto (camelCase) instead of an object (underscore)

dash/backend/src/modules/falco/service/falco.service.ts
charisprose: There's logic from the front-end to make sub-level check box(s) required when top-level box is checked. Is it sufficient?
dash/frontend/src/app/modules/private/pages/falco/falco-settings/falco-settings.component.ts#L122

dash/backend/src/modules/shared/services/email.service.ts
charisprose: I believe it returns true when the email is sent.

dash/backend/src/modules/shared/services/email.service.ts
charisprose: Should it return null or something else?

dash/frontend/src/app/modules/private/pages/falco/falco-settings/falco-settings.component.html
charisprose: The rules setting part were hidden as comment in the html file.

dash/backend/src/modules/falco/controllers/falco.controller.ts
charisprose: This is the createFalcoSettings. Was your comment for createFalcoLog? I think we already have that covered.

dash/backend/src/modules/command-line/commands/falco-email.command.ts
```diff
@@ -0,0 +1,40 @@
...
+@Injectable()
+export class FalcoEmailCommand {
+
+    constructor(private readonly loggerService: MineLoggerService,
+                private readonly falcoService: FalcoService,
+                ) {}
+
+    @Command({ command: 'schedule:falco-email', describe: 'Schedule Falco email' })
+    async get() {
+        /*const clusters = await this.clusterService.getClusters();
```
jacobbeasley, line 24: Lets add a //@TODO comment explaining that this feature hasn't been built yet, that way if someone is reading the code they aren't confused about why it is here and commented out.

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +106,48 @@ export class FalcoController {
...
+        // get all authorities from the current user
+        const currentUserAuth = currentUserAuthObj[0].authorities;
+        let isFalcoUser = this.authService.checkAuthority(currentUserAuth, [AuthorityId.FALCO]);
+        if (!isFalcoUser) {
+            throw new UnauthorizedException('Access Denied!');
+            return;
+        }
+
+        // check if falco settings rules should be applied
+        /*const rules = getFalcoSettingsRules(clusterId);
```
jacobbeasley, line 130: Needs a comment explaining why this code is here

        // @TODO: Will implement falco filters/rules settings at a later time

dash/backend/src/modules/falco/controllers/falco.controller.ts
jacobbeasley: Oh yeah, true

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -120,6 +127,115 @@ export class FalcoService {
...
+            from: this.configService.get('email.default.sender'),
+            subject: `New ${falcoSeverity} Project Falco Alert in ${falcoNamespace}`,
+            template: 'falco-log-email',
+            context: {
+                falcoLog: instanceToPlain(newFalcoLog),
+                falcoLogTags: instanceToPlain(newFalcoLog.raw["tags"]),
+                moreDetailsLink: `https://dev-m9sweeper.intelletive.com/private/clusters/${clusterId}/falco/more/${falcoId}/signature/${falcoSignature}`,
+            }
+        }).catch(e => {
+            console.log('Error sending falco email: ' + e);
```
jacobbeasley, line 162: Should use logger service

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -88,27 +106,48 @@ export class FalcoController {
...
+            throw new UnauthorizedException('Access Denied!');
+            return;
         }
+
+        // get all authorities from the current user
+        const currentUserAuth = currentUserAuthObj[0].authorities;
+        let isFalcoUser = this.authService.checkAuthority(currentUserAuth, [AuthorityId.FALCO]);
+        if (!isFalcoUser) {
+            throw new UnauthorizedException('Access Denied!');
+            return;
```
jacobbeasley, line 126: Can remove return statement. Won't be reached due to exception.

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -192,11 +195,102 @@ export class FalcoDao {
...
+
+        // if record is added, return the record
+        if (query.length > 0){
+            return query[0];
+        }
+
+        return null;
+    }
+
+        async createFalcoSetting(clusterId: number, falcoSetting: FalcoSettingDto): Promise<any>{
```
jacobbeasley, line 213: tabbing for this function seems wrong. Appears to be spaced an extra 4 spaces.

PR #31, opened 2023-01-27
Falco test fixes

charisprose:


dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -179,6 +179,11 @@ export class FalcoService {
             return; // no settings means there's no email to send!
         }
 
+        if(!falcoSetting.sendNotificationAnomaly){
+            console.log('notification is off');
```
jacobbeasley, line 183: Should use logging service

PR #33, opened 2023-01-30
Advanced Search - Remember previous input data

charisprose:


dash/frontend/src/app/modules/private/pages/image/images-list/advanced-search-dialog/advanced-search-dialog.component.ts
```diff
@@ -24,9 +24,26 @@ export class AdvancedSearchDialogComponent implements OnInit {
       cve: [''],
       onlyRunning: [true]
     });
+
+    const searchTexts = JSON.parse(localStorage.getItem('text'));
+
+    if (searchTexts[0]){
```
jacobbeasley, line 30: Instead of storing as an array of objects like this, it seems like it would be much cleaner to just store as 2 separate items in local storage. 

Also, I'm concerned the name "text" in local storage is very vague and likely to collide with other pages. 

So, I suggest the following names: 

image-search-cve
image-search-name

etc

PR #39, opened 2023-02-02
Update exception dao

charisprose:


dash/backend/src/modules/exceptions/dao/exceptions.dao.ts
```diff
@@ -193,7 +193,8 @@ export class ExceptionsDao {
 
         //console.log('getAllFilteredPolicyExceptions: ', sql.toQuery());
 
-        return await knexnest(sql).then(data => data);
+        const result = await knexnest(sql).then(data => data);
```
jacobbeasley, line 196: is the .then() necessary?

jacobbeasley: elsewhere I've seen the .then() use the plainToObject() to cast into the appropriate object class.

jacobbeasley: .then(data => plainToClass(data, ExceptionQueryDto)

PR #44, opened 2023-02-09
Upgrade KB KH KS

charisprose:


manifests/charts/dash/values.yaml
```diff
@@ -12,7 +12,7 @@ image:
 kubesec:
   registry: m9sweeper
   repository: kubesec
-  tag: latest
+  tag: 2.12.0
```
jacobbeasley, line 15: Also need to change registry do docker.io and repository to kubesec/kubesec, I think

jacobbeasley: So that we are loading from dockerhub

jacobbeasley: https://hub.docker.com/r/kubesec/kubesec/tags

manifests/charts/kube-bench/values.yaml
```diff
@@ -39,7 +39,7 @@ pod:
 
 image:
   repository: m9sweeper/kube-bench
-  tag: latest
+  tag: 0.6.11
```
jacobbeasley, line 42: I also want to see if perhaps we can stop having our own custom kube-bench image. Touch base with Grant on whether we could change the entrypoints in our kube-bench helm chart to install curl/wget or whatever

manifests/charts/dash/values.yaml
jacobbeasley: We can also remove this file - https://github.com/m9sweeper/m9sweeper/blob/main/dash/kubesec/Dockerfile

manifests/charts/kube-bench/values.yaml
jacobbeasley: https://github.com/m9sweeper/m9sweeper/blob/main/manifests/addons/kube-bench/Dockerfile

So we'd have to change the entrypoints to be prefixed with...

apk update && apk add curl &&

jacobbeasley: Then we could remove https://github.com/m9sweeper/m9sweeper/blob/main/manifests/addons/kube-bench/Dockerfile

jacobbeasley: https://github.com/m9sweeper/m9sweeper/blob/main/manifests/charts/kube-bench/templates/cronjob-ack.yaml#L62 - you can see how each job in the templates right now runs kube-bench in different ways. Instead of having our own image in order to install curl, we could just have it install curl when it boots up.

manifests/charts/kube-bench/templates/cronjob.yaml
```diff
@@ -59,7 +59,7 @@ spec:
               image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
               imagePullPolicy: {{ .Values.image.pullPolicy }}
               command: [ "/bin/sh" ]
-              args: [ "-c", "kube-bench --json | curl -s -H \"Content-Type: application/json\" -X POST --data-binary @- {{ .Values.reportsurl }}" ]
+              args: [ "-c", "apk update && apk add curl && kube-bench --json | curl -s -H \"Content-Type: application/json\" -X POST --data-binary @- {{ .Values.reportsurl }}" ]
```
jacobbeasley, line 62: @charisprose  I believe we have to change like 20+ other files in this same directory.

jacobbeasley: There are a bunch of other job/cron job specs that will need to be updated in the same way. If you just look at the other files in this folder it will become clear.

manifests/charts/kube-bench/values.yaml
```diff
@@ -38,8 +38,13 @@ pod:
     "helm.sh/hook-delete-policy": before-hook-creation
 
 image:
+<<<<<<< Updated upstream
```
jacobbeasley, line 41: Looks like a bad merge

manifests/charts/kube-bench/values.yaml
```diff
@@ -38,8 +38,13 @@ pod:
...
   repository: m9sweeper/kube-bench
-  tag: latest
+  tag: 0.6.11
+=======
+  repository: aquasec/kube-bench
+  tag: v0.6.11
+>>>>>>> Stashed changes
```
jacobbeasley, line 47: bad merge

jacobbeasley:
@charisprose  I see backend/package-lock was changed but I also don't see any changes to package.json. Maybe that was committed by mistake.

dash/backend/src/modules/auth/auth.module.ts
```diff
@@ -13,8 +13,6 @@ import {LdapFactory} from './factories/LdapFactory';
 import LoggedInUserFactory from '../../providers/LoggedInUserFactory';
 import {ResetPasswordService} from './services/reset-password.service';
 import {ResetPasswordDao} from './dao/reset-password.dao';
-import {AppSettingsService} from '../settings/services/app-settings.service';
```
jacobbeasley, line 16: Why are we removing this?

jacobbeasley: Is it just dead code?

manifests/charts/kube-bench/templates/cronjob.yaml
charisprose: Yup, I mentioned this in the chat. Will update the rest on Monday.

manifests/charts/kube-bench/values.yaml
charisprose: Not sure why it appeared like this. I copied and pasted the line. Maybe it didn't like the format. I'll fix this.

dash/backend/src/modules/auth/auth.module.ts
charisprose: When I tried to boot up locally, I was prompted to remove so (not allowed to be in certain modules). It might be from the nestjs upgrade.

manifests/charts/kube-bench/values.yaml
jacobbeasley: https://hub.docker.com/r/aquasec/kube-bench/tags
Need the v

PR #45, opened 2023-02-10
Falco bulkhead

GrantWK:
Tested and confirmed this is routing to dash-falco


manifests/charts/dash/values.yaml
```diff
@@ -20,6 +20,10 @@ kubesec:
     password: ""
     secretRef: ""
 
+falco:
```
jacobbeasley, line 23: Lets add a comment explaining as well as update the advanced install section of the docs

PR #63, opened 2023-03-06
Fix Falco date/time sort

charisprose:


dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -69,10 +69,10 @@ export class FalcoDao {
                     );
                     break;
                 case 'Date Desc':
-                    query = query.orderBy([{column: 'calendar_date', order: 'desc'}]);
+                    query = query.orderBy([{column: 'creation_timestamp', order: 'desc'}]);
```
jacobbeasley, line 72: One thing to check - how does this query perform? Does it hit indexes or anything?

PR #72, opened 2023-03-11
added gcr.io registry 

vellankikoti:


manifests/charts/m9sweeper/values.yaml
```diff
@@ -64,7 +64,8 @@ dash:
           aliases:
             - index.docker.io
           login_required: false
-
+        - name: gcr.io
+          hostname: registry.k8s.io
```
jacobbeasley, line 68: Should we set login-required to false by default?

vellankikoti: yes Jake, I missed it sorry adding it

jacobbeasley: Thanks!

PR #74, opened 2023-03-13
Update timestamp variable

charisprose:


dash/frontend/src/app/core/dialogues/show-json-data/show-json-data.component.ts
```diff
@@ -27,7 +27,7 @@ export class ShowJsonDataComponent implements OnInit {
   displayedColumns = ['calendarDate', 'namespace', 'pod', 'image', 'message'];
   namespace = this.data.content.namespace;
   date = this.data.content.calendarDate;
-  dateTime  = new Date(this.date);
+  dateTime  = new Date(+(this.data.content.timestamp));
```
jacobbeasley, line 30: Is this guaranteed to be populated? If not, we may want to handle cases where it is not populated elegantly.

jacobbeasley: Well, I'm sure content won't be null, but I wonder if timestamp might ever be null. If it were null, what would the output of the code do?

jacobbeasley: IE - would this code potentially crash the application? I'm not sure, but you should check since its operating on that field: 

    dateTime  = new Date(+(null));

PR #77, opened 2023-03-29
optimizing count queries and indexes for project falco

jacobbeasley:


jacobbeasley:
Count queries in postgresql are slow... so this sidesteps the issue by either constraining to a specific set of days or just not counting at all.

PR #88, opened 2023-04-21
Fix Falco download button

charisprose:


dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -59,15 +59,13 @@ export class FalcoController {
...
-        @Query('eventId') eventId?: number
-    ): Promise<{ logCount: number, list: FalcoDto[]}>
+    ): Promise<{ logCount: number, list: FalcoDto[],  csvLogList: FalcoDto[]}>
     {
         // default values - seems to be filling with NaN instead of null which causes the defaults in Dao to not work right
         if (!limit) limit = null;
         if (!page) page = null;
-        if (!eventId) eventId = null;
 
-        return this.falcoService.getFalcoLogs(clusterId, limit, page, priorities, orderBy, startDate, endDate, namespace, pod, image, signature, eventId);
```
jacobbeasley, line 70: Why was this parameter removed? Is it ever used anywhere?

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +101,32 @@ export class FalcoDao {
...
         }
 
+        // Filtered list for csv - limit to 1000 logs
+        const filteredCsvLoglist = await query.limit(1000).then(data => {
+            return plainToInstance(FalcoDto, data);
+        });
+
         if (limit){
-            query = query.limit(limit).offset(page * limit)
+            query = query.limit(limit).offset(page * limit) // limit: page size
```
jacobbeasley, line 110: Where precisely is it defaulting the limit and page number?

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -78,13 +77,29 @@ export class FalcoService {
...
+        // retrieve filtered falco logs and use the query results to build the csv
+        const queryResp = await this.falcoDao.getFalcoLogs(clusterId, limit, page, priorities, orderBy, startDate, endDate, namespace, pod, image, signature);
         const result = [this.csvService.buildLine(['Date', 'Namespace', 'Pod',
             'Image', 'Priority', 'Message'])];
 
-        for (let i = 0; i < queryResponse.logCount; i++) {
-            const falcoCol = queryResponse.fullList[i];
+        // limit to 1000 or less logs
+        let numOfLogs = 0;
+        queryResp.csvLogList.length <= 1000 ? numOfLogs = queryResp.csvLogList.length : numOfLogs = 1000;
```
jacobbeasley, line 100: A more common way you see people do this is like this.. 

const numOfLogs = queryResp.csvLogList.length <= 1000 ? queryResp.csvLogList.length : 1000

Also, enforcing the limit here is not ideal because its enforcing it AFTER the query has already happened. Better to enforce the limit higher up in the code prior to the SQL query occurring, that way if they pass in a limit of 100000000000 then it will never actually run against the postgresql server with that limit.

dash/frontend/src/app/modules/private/pages/falco/falco-events-list/falco-events-list.component.ts
```diff
@@ -86,7 +86,11 @@ export class FalcoEventsListComponent implements OnInit {
     this.limit = pageEvent.pageSize;
     this.page = pageEvent.pageIndex;
     this.setLimitToLocalStorage(this.limit);
-    this.getEvents();
+
+    console.log('limit now', this.limit);
```
jacobbeasley, line 90: We can comment these out

jacobbeasley: no need to log here I think

dash/backend/src/modules/falco/controllers/falco.controller.ts
charisprose: As I was making changes to getFalcoLogs, I realized that the eventId was never used in the query, not event passed from the frontend. Event id is essentially the primary id. We already have the desc/ asc filter to cover the sorting. I think it's safe to remove something we don't use.

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -78,13 +77,29 @@ export class FalcoService {
...
+                       endDate?: string,
+                       namespace?: string,
+                       pod?: string,
+                       image?: string,
+                       signature?: string,
+                       ): Promise<FalcoCsvDto> {
+
+        // retrieve filtered falco logs and use the query results to build the csv
+        const queryResp = await this.falcoDao.getFalcoLogs(clusterId, limit, page, priorities, orderBy, startDate, endDate, namespace, pod, image, signature);
         const result = [this.csvService.buildLine(['Date', 'Namespace', 'Pod',
```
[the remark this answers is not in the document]
charisprose: Since we're limiting the results in the getFalcoLogs() dao from the other comment, do we still need to put a limit here?

dash/backend/src/modules/falco/service/falco.service.ts
charisprose: I'll keep the falco dao line 105 and the logic here so we can still get download logs up to 1000. "Limit" came through from the frontend as the page size. If we use limit in both of these places, we'll get only 20, 50 or 100 download logs. Let me know what you think.

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +104,31 @@ export class FalcoDao {
             query.where('anomaly_signature', signature);
         }
 
-        if (limit){
-            query = query.limit(limit).offset(page * limit)
-        }
+        // Filtered list for csv - limit to 1000 logs
+        const filteredCsvLoglist = await query.limit(1000).then(data => {
```
jacobbeasley, line 108: Instead of hardcoding the limit, it should use the variable from line 34

jacobbeasley: Also, where is offset being set?

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +104,113 @@ export class FalcoDao {
...
+
+        if (startDate) {
+            query = query.andWhere('calendar_date', '>=', startDate);
+        }
+        if (endDate) {
+            query = query.andWhere('calendar_date', '<=', endDate);
+        }
+
+        if (namespace) {
+            query = query.whereRaw(`namespace LIKE ?`, [`%${namespace.trim()}%`]);
```
jacobbeasley, line 196: whenever using raw queries, we need to be careful to prevent sql injection. I think that in this case, sql injection would be possible. We should use regex to ensure that namespace is alphanumeric OR dashes (capital or lowercase letters, numbers, or dashes)

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +104,113 @@ export class FalcoDao {
...
+        }
+
+        if (pod) {
+            query = query.whereRaw(`container LIKE ?`, [`%${pod.trim()}%`]);
```
jacobbeasley, line 200: same here

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +104,113 @@ export class FalcoDao {
...
+        }
+
+        if (image) {
+            query = query.whereRaw(`image LIKE ?`, [`%${image.trim()}%`]);
```
jacobbeasley, line 204: same here

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -78,13 +77,28 @@ export class FalcoService {
...
+
+        // retrieve filtered falco logs and use the query results to build the csv
+        const queryResp = await this.falcoDao.getFalcoCsvLogs(clusterId, priorities, orderBy, startDate, endDate, namespace, pod, image,);
         const result = [this.csvService.buildLine(['Date', 'Namespace', 'Pod',
             'Image', 'Priority', 'Message'])];
 
-        for (let i = 0; i < queryResponse.logCount; i++) {
-            const falcoCol = queryResponse.fullList[i];
+        // limit to 1000 or less logs
+        const numOfLogs = queryResp.csvLogList.length <= 1000 ? queryResp.csvLogList.length : 1000;
```
jacobbeasley, line 99: Now that we are limiting the resultset size properly in the DAO, you can simplify this code to just a few lines: 

    const result = queryResp.csvLogList.map(falcoCol => this.csvService.buildLine([
                String(falcoCol.calendarDate),
                String(falcoCol.namespace),
                String(falcoCol.container),
                String(falcoCol.image),
                String(falcoCol.level),
                String(falcoCol.message),
            ]));

dash/frontend/src/app/core/services/falco.service.ts
```diff
@@ -34,7 +34,7 @@ export class FalcoService {
   ) {}
 
   getFalcoLogs(clusterId: number, options?: FalcoLogOptions
-  ): Observable<IServerResponse<{logCount: number, list: IFalcoLog[]}>> {
+  ): Observable<IServerResponse<{logCount: number, list: IFalcoLog[], csvLogList: IFalcoLog[]}>> {
```
jacobbeasley, line 37: I think this change may not be needed

dash/backend/src/modules/falco/dao/falco.dao.ts
jacobbeasley: For example, if someone passed in a namespace of the following, it would be bad: 

'; DELETE FROM user; SELECT * from project_falco_logs where ''='

jacobbeasley: Did some googling and the way you are using string interpolation is actually correct. 

https://stackoverflow.com/questions/50327070/how-do-i-escape-in-knex-where-like-query

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -102,26 +104,113 @@ export class FalcoDao {
...
+        }
+
+        // Filtered list for csv - limit to 1000 logs
+        const filteredCsvLoglist = await query.limit(1000).then(data => {
+            return plainToInstance(FalcoDto, data);
```
jacobbeasley, line 209: fun fact for the future - these can just be...

.then(d -> plainToInstance(FalcoDto, d))

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -78,22 +77,34 @@ export class FalcoService {
...
+        // retrieve filtered falco logs and use the query results to build the csv
+        const queryResp = await this.falcoDao.getFalcoCsvLogs(clusterId, priorities, orderBy, startDate, endDate, namespace, pod, image,);
+        let result = [this.csvService.buildLine(['Date', 'Namespace', 'Pod',
             'Image', 'Priority', 'Message'])];
 
-        for (let i = 0; i < queryResponse.logCount; i++) {
-            const falcoCol = queryResponse.fullList[i];
-            result.push(this.csvService.buildLine([
+        // limit to 1000 or less logs from dao
+            queryResp.csvLogList.map(falcoCol => result.push(this.csvService.buildLine([
```
jacobbeasley, line 99: If you are not going to use the .map(), then I'd just change .map() to .forEach()

PR #91, opened 2023-05-02
WIP: Implementing Azure OAuth2.0

sforkani:


dash/backend/src/main.ts
```diff
@@ -177,6 +177,8 @@ async function bootstrap() {
     }
   });
 
+  app.use(urlencoded({ extended: true }));
```
jacobbeasley, line 180: What is this for?

jacobbeasley:
I'm going to merge so we can start testing, but lets do another PR to fill in reasonable defaults for the URLs for Google/Microsoft OAuth. @sforkani

PR #93, opened 2023-05-15
Fix proxy webhook install

brandan-schmitz:
Fixes an issue with sed in the auto-setup.sh script. SED does not recognize `-I` as a valid argument on the linux versions and it also does not work with the extra `''` in the command. This fixes this by checking if the OS is linux or not and running a different version of the command on linux vs maxOS machines to account for the differences.

Also adds `set -e` to allow the script to return a error code if a portion fails so that when used in pipelines it allows a stage to fail upon failure of the script.

### DIFF
--- a/dash/backend/scripts/proxy-webhook/auto-setup.sh
+++ b/dash/backend/scripts/proxy-webhook/auto-setup.sh
@@ -58,8 +58,14 @@ cp initialize_proxy.yaml init.yaml
 cp openssl.cnf ssl.cnf
 
 # Change the namespace in the initialize file and openssl configuration
-sed -I '' -e 's/m9sweeper-system/'$namespace'/g' init.yaml
-sed -I '' -e 's/m9sweeper-system/'$namespace'/g' ssl.cnf
+platform=$(uname)
+if [[ "$platform" != 'Linux' ]]; then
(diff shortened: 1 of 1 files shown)


dash/backend/scripts/proxy-webhook/auto-setup.sh
```diff
@@ -55,16 +58,22 @@ cp initialize_proxy.yaml init.yaml
 cp openssl.cnf ssl.cnf
 
 # Change the namespace in the initialize file and openssl configuration
-sed -I '' -e 's/m9sweeper-system/'$namespace'/g' init.yaml
-sed -I '' -e 's/m9sweeper-system/'$namespace'/g' ssl.cnf
+platform=$(uname)
+if [[ "$platform" != 'Linux' ]]; then
+    sed -i "s/m9sweeper-system/$namespace/g" init.yaml
```
jacobbeasley, line 63: I noticed this is now lower case. Not sure if that was intended or not, otherwise I'll merge it.

brandan-schmitz: That is correct, `-I` with the capital I is not a valid argument in linux and in the help command for macOS it even shows to use a lowercase i.

PR #101, opened 2023-05-31
Fix for double scan results being sent for images added with + button

kyle-intelletive:


kyle-intelletive:
Quick summary:

Trawler always sent tthe scan results.

If the image hash didn't match Trawler's pull it ALSO sent a failed scan image. When adding via + button, the hash was set to a generated TMP_.... one, and that casused that error.

Trawler now sends the error one if applicable, or the scan result otherwise, and ignores that case if the hash is a TMP_... one

PR #103, opened 2023-06-06
Add domain validation to OAuth, add Oauth docs

jshoberg:


docs/content/en/docs/Concepts/SignOnMethods/oauth2.md
```diff
@@ -0,0 +1,80 @@
+---
+title: "OAuth2 Login"
+linkTitle: "OAuth2 Login"
+weight: 2
+description: >
+  How to configure user login using OAuth2
+---
+
+M9sweeper currently supports two implementations of Oauth2, through Google and Microsoft Azure. To add one of these
```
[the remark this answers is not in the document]
jacobbeasley: Might mention we support really any oauth2 implementation, but we've tested it to work with both of them

docs/content/en/docs/Concepts/SignOnMethods/oauth2.md
```diff
@@ -0,0 +1,80 @@
...
+select the OAUTH2 Auth Type. After, select your desired implementation from the Provider Type dropdown, with
+implementation details found below.
+
+Make sure to give the sign-on method a unique name. Note that the Auth Name selected will be what users see when
+selecting a login method.
+
+## Google Oauth2
+
+Using Google OAuth requires first setting up credentials through Google's cloud platform console. Instructions on
+setting up necessary credentials can be found through [Google's official docs](https://support.google.com/cloud/answer/6158849?hl=en).
```
jacobbeasley, line 20: excellent!

docs/content/en/docs/Concepts/SignOnMethods/oauth2.md
```diff
@@ -0,0 +1,80 @@
...
+
+After creating your credentials, you will need to take the Client ID and Client Secret values generated by Google and
+add them to their respective fields in the configuration form.
+
+![../img.png](../img.png)
+
+### Access Token Uri and Authorization Uri
+
+These fields refer to the endpoints which the app will reach out to during the authorization process. They are
+automatically filled with the default values used by Google, and you shouldn't need to modify them yourself
```
jacobbeasley, line 31: needs period at the end

jacobbeasley: should not

docs/content/en/docs/Concepts/SignOnMethods/oauth2.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Access Scopes
+
+A comma separated list of the fields the app will request access to when querying data from a user. This is automatically
+filled with the information M9sweeper needs to access, and you shouldn't need to modify it yourself.
```
jacobbeasley, line 36: should not - usually you avoid contractions in technical writing

dash/frontend/src/app/modules/private/pages/external-auth-configuration/external-auth-configuration-create/external-auth-configuration-create.component.html
```diff
@@ -67,6 +67,13 @@ <h2 mat-dialog-title class="subnav-title">
               <input formControlName="oauthScopes" matInput>
             </mat-form-field>
           </p>
+          <p>
+            <mat-form-field>
+              <mat-label>Allowed User Domains (comma separated)</mat-label>
```
jacobbeasley, line 72: is this a required field? (if not, maybe it should be to prevent insecure usage)

dash/frontend/src/app/modules/private/pages/external-auth-configuration/external-auth-configuration-create/external-auth-configuration-create.component.html
```diff
@@ -18,11 +18,11 @@ <h2 mat-dialog-title class="subnav-title">
...
     </p>
     <p *ngIf="oauthAuthActivated">
-      <mat-form-field (selectionchange)="onOAuthProviderTypeChange($event)">
+      <mat-form-field>
         <mat-label>Provider Type</mat-label>
-        <mat-select formControlName="providerType">
-          <mat-option value="GOOGLE">GOOGLE</mat-option>
-          <mat-option value="AZURE">Azure</mat-option>
+        <mat-select formControlName="providerType" (selectionChange)="onOAuthProviderTypeChange($event)">
+          <mat-option [value]="'GOOGLE'">Google</mat-option>
```
jacobbeasley, line 24: Why was this change needed?

dash/backend/src/modules/auth/services/oauth2/google-oauth2.service.ts
```diff
@@ -43,9 +44,15 @@ export class GoogleOauth2Service extends Oauth2AuthProvider {
 
     userProfile.authorities = [];
 
+    // Check access rights for profile
     const userAuthority = new UserAuthority();
-    userAuthority.id = 3;
-    userProfile.authorities.push(userAuthority);
+    const emailDomain = userProfile.email.split('@').pop().toLowerCase().trim();
+    if (oAuth2Config.allowedDomains.includes(emailDomain)) {
```
jacobbeasley, line 50: may want to make sure the allowed domains are lowercased, too, if its not doin gthat already

dash/backend/src/modules/auth/services/oauth2/azure-oauth2.service.ts
```diff
@@ -35,9 +37,15 @@ export class AzureOauth2Service extends Oauth2AuthProvider {
...
     userProfile.authorities = [];
 
+    // Check access rights for profile
     const userAuthority = new UserAuthority();
-    userAuthority.id = 3;
-    userProfile.authorities.push(userAuthority);
+    const emailDomain = userProfile.email.split('@').pop().toLowerCase().trim();
+    if (oAuth2Config.allowedDomains.includes(emailDomain)) {
+      userAuthority.id = AuthorityId.READ_ONLY;
+      userProfile.authorities.push(userAuthority);
```
jacobbeasley, line 45: we'd only want to modify user authorities if its being created right now.

jacobbeasley: if it already existed it would use whatever user profiles already existed

dash/frontend/src/app/modules/private/pages/external-auth-configuration/external-auth-configuration-create/external-auth-configuration-create.component.html
jshoberg: It's not, I can set it to be

dash/frontend/src/app/modules/private/pages/external-auth-configuration/external-auth-configuration-create/external-auth-configuration-create.component.html
jshoberg: The selection change was placed on the wrong tag previously, it needed to be moved

The capitalization on Google was changed for consistency with the Azure option

dash/backend/src/modules/auth/services/oauth2/google-oauth2.service.ts
jshoberg: The allowed domains are validated to be lowercased in the external auth config dto

```
@IsNotEmpty()
@IsString({each: true})
@IsLowercase({each: true})
allowedDomains: string[];
```

And when creating a config, the domains are changed to lowercase before passing it to the backend. Is there anywhere else they should be validated for that?

dash/backend/src/modules/auth/services/oauth2/google-oauth2.service.ts
```diff
@@ -43,9 +44,15 @@ export class GoogleOauth2Service extends Oauth2AuthProvider {
...
+      userAuthority.id = AuthorityId.READ_ONLY;
+      userProfile.authorities.push(userAuthority);
```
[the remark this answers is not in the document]
jshoberg: Right now, the OAuth flow pulls information from the OAuth provider and creates a default user, then it checks to see if a user with that email already exists in the system; if that user exists, it ignores the default user and loads in the existing user (assuming the user is active, it throws an error if the user exists but is inactive), and if the user doesn't exist it creates a new user with the information from OAuth. See the redirectCallbackHandler function [here](https://github.com/m9sweeper/m9sweeper/blob/b8289647c28681680b09af5246dcb196f8b8c5a4/dash/backend/src/modules/auth/controllers/redirectable-login.controller.ts#L69)

It probably would make more sense to have that logic in the service layer, I could move it there if you want? It would probably make the code easier to follow.

dash/backend/src/modules/auth/services/oauth2/google-oauth2.service.ts
jacobbeasley: That works!

docs/content/en/docs/Concepts/SignOnMethods/oauth2.md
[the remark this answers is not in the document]
jshoberg: Should I say that you can overwrite the endpoints to access other oauth2 implementations? Or just leave it at the mention that we've tested those two?

PR #104, opened 2023-06-08
Ensure active policies have at least one active scanner

jshoberg:


dash/backend/src/modules/policy/services/policy-service.ts
```diff
@@ -93,5 +94,15 @@ export class PolicyService {
         return await this.auditLogService.calculateMetaData(previous, updated, 'Policy');
     }
 
+    validateActivePolicyScanners(policy: PolicyDto, scanners: ScannerDto[]): boolean {
+        // Checks if an active policy has at least one active scanner
+        if (policy.enabled) {
```
jacobbeasley, line 99: Might be able to simplify using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

jacobbeasley: return !policy.enabled || scanners.some(...)

jacobbeasley: return !policy.enabled || scanners.some(scanner?.enabled);

jshoberg: Haven't run into that function before, thanks

PR #108, opened 2023-06-08
Update falco output URL and baseUrl default

brandan-schmitz:
Updates the command shown to users to install m9sweeper to fix the http output URL. Also changes the baseUrl in the main chart to include the full namespace so falco can access it since its installed in a different namespace if not using an ingress.

### DIFF
--- a/manifests/charts/m9sweeper/values.yaml
+++ b/manifests/charts/m9sweeper/values.yaml
@@ -29,7 +29,7 @@ global:
   # Provide a secret string that will be used to sign JWT tokens, it is highly recommended that you set this to your own secret
   jwtSecret: "this-is-not-secret"
   # This URL will be used in email templates to reference a http link to Dash
-  baseUrl: "http://m9sweeper-dash:3000"
+  baseUrl: "http://m9sweeper-dash.m9sweeper-system.svc:3000"
   # Provide a secret string that will be set as the trawler service account's API key.
   # It will also be passed to trawler so that it can auth with Dash. You will not need it
   # after the initial installation as it will be loaded into the DB.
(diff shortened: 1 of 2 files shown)


manifests/charts/m9sweeper/values.yaml
```diff
@@ -29,7 +29,7 @@ global:
   # Provide a secret string that will be used to sign JWT tokens, it is highly recommended that you set this to your own secret
   jwtSecret: "this-is-not-secret"
   # This URL will be used in email templates to reference a http link to Dash
-  baseUrl: "http://m9sweeper-dash:3000"
+  baseUrl: "http://m9sweeper-dash.m9sweeper-system.svc:3000"
```
jacobbeasley, line 32: So, few thoughts on this: 
 - The postfix ".svc" is actually dynamic (it can be overridden in kubernetes), but I suppose that's tolerable
  - technically they may not be installing our app in m9sweeper-system, though I suppose that is the default. 
  
  I wonder if we could use ${Release.Namespace}
  
  https://helm.sh/docs/chart_template_guide/builtin_objects/

jacobbeasley: I'll merge either way but give that some thought tomorrow.

brandan-schmitz: With that being in the default values file it would be expected they would update the m9sweeper-system from there if they are installing in a different namespace. That would be a more advanced install since all of our documentation is based around them using the m9sweeper-system namespace.

As for the svc part yes it can be overridden but again if that has happened that is something beyond the scope of m9sweeper and is in fact a fairly uncommon occurrence so that would be expected for them to be able to account for that in adjusting their values.

PR #109, opened 2023-06-08
Use init containers

brandan-schmitz:
Use init containers to ensure that postgresql is accepting connections before starting dash pod and the dash init job. Should prevent some race condition issues in some installs.


manifests/charts/dash/templates/deployment.yaml
```diff
@@ -26,6 +26,25 @@ spec:
         {{- include "dash.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccount: dash-sa
+      initContainers:
+        - name: wait-for-postgresql
+          image: postgres:11.10
```
jacobbeasley, line 31: Need to make this image name dynamic like other images because if they are deploying on-premise or in a firewalled environment they won't be able to pull from dockerhub (perhaps use same image as postgres image variable)

manifests/charts/dash/templates/deployment.yaml
```diff
@@ -26,6 +26,25 @@ spec:
...
+          imagePullPolicy: IfNotPresent
+          command:
+            - 'sh'
+            - '-c'
+            - 'until pg_isready; do echo "Waiting for database...""; sleep 2; done;'
```
jacobbeasley, line 36: nice

manifests/charts/dash/templates/deployment.yaml
brandan-schmitz: We would need to introduce a new variable such as postgresInitImage.repository and postgresInitImage.tag in the dash chart itself if we want to do this. With the postgres being in a sibling level dependency chart under the m9sweeper install we cannot get the value from it directly as it is not in the scope for when it is parsing the values for the dash portion. We also would not want to do that because if they disable the bundled postgres installation it again would not be able to find the image.

I can add a section to the dash chart though for the postgres init container.

jacobbeasley: Yeah, just need some way for them to override it. Maybe its a separate variable, then. 

The other option would be to use the dash image and just check postgresql is online using nodejs instead of introducing a new iamge.

jacobbeasley: One thing we've been doing is trying to reduce image count. Makes deploying it on-premise way easier when we don't have more images than we need.

brandan-schmitz: It now includes a new CLI command to check the database status and uses the dash image already defined.

PR #112, opened 2023-06-12
Add Azure Container Registry support to m9sweeper

jshoberg:


trawler/src/main/java/io/m9sweeper/trawler/scanners/Trivy.java
```diff
@@ -117,6 +117,18 @@ public void runScan() throws Exception {
 
             // Export the location of this file so that Trivy can utilize it
             trivyScanCommandBuilder.append("export GOOGLE_APPLICATION_CREDENTIALS='").append(gcrAuthFile.getAbsolutePath()).append("'; ");
+        } else if (registry.getAuthType().equals("AZCR")) {
```
jacobbeasley, line 120: not a blocker, but wanted to mention in Java people oftentimes flip it around like this: 


"AZR".equals(registry.getAuthType())

That way if auth type is null/not set, it doesn't throw a null pointer exception

jacobbeasley: See also: https://en.wikipedia.org/wiki/Yoda_conditions

trawler/src/main/java/io/m9sweeper/trawler/scanners/Trivy.java
```diff
@@ -117,6 +117,18 @@ public void runScan() throws Exception {
...
+            // Azure Container Registry images are accessed with a service principal set up beforehand. Trawler only needs to
+            // export the Client ID, Secret, and Tenant ID of the service principal to allow Trivy to connect to it
+            Map<String, Object> authDetails = (Map<String, Object>) registry.getAuthDetails();
+
+            String clientId = authDetails.get("azureClientId").toString();
```
jacobbeasley, line 125: I believe there is a getOrDefault() that you can use that is null safe

trawler/src/main/java/io/m9sweeper/trawler/scanners/Trivy.java
```diff
@@ -117,6 +117,18 @@ public void runScan() throws Exception {
...
+            String clientSecret = authDetails.get("azureClientSecret").toString();
+            String tenantId = authDetails.get("azureTenantId").toString();
+
+            trivyScanCommandBuilder.append("export AZURE_CLIENT_ID='").append(escapeXsi(clientId)).append("'; ");
```
jacobbeasley, line 129: nice!

PR #119, opened 2023-06-22
Docker compose

kyle-intelletive:
### Runs & networks these services:
- dash
- dash-init jobs
- trawler
- kubesec
- rabbit mq
- postgres (with persistent storage)
### Other
- Inculdes Readme & sample env
- created functionality to enable running the cronjobs via nestjs/schedule

## nestjs-command removal
- removed nestjs-command package.
- refactored npm run commands
- updated readme to include that building is necessary to run the CLI commands necessary
- refactored commands to return booleans for success or failure.
- Created "controller" for running the commands

## Misc changes:
- Moved .dockerignore to be in same directory as Dockerfile
- Upgraded kubernetes-client-node. Changes made:
  - V1Event -> CoreV1Event
  - jsYaml started returning 'unknown' instead of 'any' converted to any to work as behavior and make compiler not complain
  - Made cluster sync gracefully handle being called for clusters that don't exist
  - Kubesec url is configurable (defaults to localhost 8080 for backwards compatibility)


docker-compose/.env.sample
```diff
@@ -0,0 +1,57 @@
+##############################
+### User Init Job Settings ###
+##############################
+# These variables will be used by the init jobs the first time you boot up with a fresh DB
+SUPER_ADMIN_EMAIL=
+SUPER_ADMIN_PASSWORD=
+INITIAL_REGISTRIES_JSON="eyJyZWdpc3RyaWVzIjpbeyJuYW1lIjoiRG9ja2VyaHViIiwiaG9zdG5hbWUiOiJkb2NrZXIuaW8iLCJhbGlhc2VzIjpbImluZGV4LmRvY2tlci5pbyJdLCJsb2dpbl9yZXF1aXJlZCI6ZmFsc2V9LHsibmFtZSI6IkdpdGh1YiBDb250YWluZXIgUmVnaXN0cnkiLCJob3N0bmFtZSI6ImdoY3IuaW8iLCJsb2dpbl9yZXF1aXJlZCI6ZmFsc2V9LHsibmFtZSI6Ikt1YmVybmV0ZXMgQ29udGFpbmVyIFJlZ2lzdHJ5IiwiaG9zdG5hbWUiOiJyZWdpc3RyeS5rOHMuaW8iLCJsb2dpbl9yZXF1aXJlZCI6ZmFsc2V9XX0="
```
jacobbeasley, line 7: I think we changed this

jacobbeasley: maybe

docker-compose/.env.sample
```diff
@@ -0,0 +1,57 @@
...
+
+################
+### API KEYS ###
+################
+# The first time you boot up against a fresh DB, these will be used to configure the api keys for these service accounts
+# After that point, they will be used in the relevant pods for authentication with the API
+TRAWLER_API_KEY=111111
```
jacobbeasley, line 14: For sample maybe we default to blank

docker-compose/.env.sample
kyle-intelletive: We only changed how it is added to the value files.

The helm chart takes a pretty list in YAML [1], but then as part of the templating [2], it converts it to a JSON and b64 encodes that to end up with this kind of string. (This one contains the same data that you would get by using the default)

[1] https://github.com/m9sweeper/m9sweeper/blob/main/manifests/charts/dash/values.yaml#L59
[2] https://github.com/m9sweeper/m9sweeper/blob/main/manifests/charts/dash/templates/job-init.yaml#L36

docker-compose/.env.sample
kyle-intelletive: I can do that, and update the README to require ading those.

My logic for providing initial values is that since this sample environment is intended to be used for local development, the security wouldn't be a big concern. If they don't fill it in and boot up, then it will be randomly generated, and would probably be confusing to people.

PR #120, opened 2023-06-22
WIP - Metrics & Monitoring

becky-intelletive:


dash/backend/src/modules/metrics/services/m9sweeper_cron_jobs.service.ts
```diff
@@ -1,41 +1,52 @@
...
         private readonly imageService: ImageService,
         private readonly exceptionsService: ExceptionsService,
     ) {}
 
     /**
-     * Recompile M9sweeper Metrics every 5 minutes for Prometheus metrics
+     * Recompile M9sweeper Metrics every 30s for Prometheus metrics
      */
-    @Cron('0 */5 * * * *')
+    @Cron('* * 01 * * *')
```
jacobbeasley, line 22: I think we discussed getting rid of this in leau of the new metrics API you are using. 

So then we'd have 2 metrics APIs: 
 - one that provides instance-level metrics (cpu usage, ram usage, response times, request volume, error rate)
 - One that provides env level details

So then this one would be supplanted by the new one that provides env level details.

becky-intelletive: Yes, that's the plan in the next PR :)

becky-intelletive: (Sorry in this PR not the next one - this is just a work in progress)

jacobbeasley:
Sounds like this PR is no longer hte right - has been broken into 2 separate PRs

PR #123, opened 2023-06-23
Falco filter rules

kyle-intelletive:
Falco rules implemented

## UI / CRUD
- CRUD endpoints & table added for falco rules
- Lists rules on main Falco settings page
- Adding a rule or clicking 'edit' for an existing rule will open a dialog where the user can input:
  - The action (silence or ignore)
  - *namespace (dropdown populated from cluster)
  - *Falco rule
  - *image  
  - *These three fields are all individually optional, but at least one must be chosen
- Can soft delete rules

## Enforcing
- Check rules when accepting falco events
- Finds the oldest applicable rule that matches (see below) the event
- If applicable rule with the 'ignore' action, the return successfully without saving or doing anything to the event.
- If applicable rule with the 'silence' action, then a flag is passed so the the email sending is skipped after the event is saved
- If no rule that applies is found, behave as before

## Matching
A rule is considered a match for an event if all 3 of the below are true:
1. One of the following is true:
   1.  The rule had no configured namespace
   2. The rule was configured with the same namespace as the event's namespace field (event.outputFields.k8sNamespaceName)
2. One of the following is true
   1. The rule had no configured falco rule
   2. The rule's falco rule, and the event's rule are the same (event.rule)
3. One of the following is ture
  1. The rule had no configured image
  2. The rule's image and the events image are the same (event.outputFields?.containerImageRepository)

Note: that for all of these fields, it must be an exact case-sensitive match with the current implementation.

## Misc
- Styles on the falco settings page were very weird, did some clean up (did not touch functionality)
- linter did some work on some of the pages I touched.


dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -154,33 +160,27 @@ export class FalcoController {
...
-                }
-            }
-        */
-
-        // Saved new falco log
-        return await this.falcoService.createFalcoLog(clusterId, falcoLog);
+        const action = await this.falcoService.checkRules(clusterId, falcoLog);
+        // An ignore rule applied,
+        if (action === FalcoRuleAction.Ignore) {
+            this.logger.log('Ignoring Falco Event');
```
jacobbeasley, line 178: Maybe should include some details about the event, otherwise you'll just know that "some event" was ignored but nothing about which event

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -192,4 +192,44 @@ export class FalcoController {
...
+      @Body() rule: FalcoRuleDto
+    ): Promise<FalcoRuleDto> {
+        if (rule?.clusterId !== clusterId || ruleId !== rule?.id) {
+            throw new HttpException({ message: 'rule or cluster id in path does not match the body.' }, HttpStatus.BAD_REQUEST);
+        }
+        return this.falcoService.updateFalcoRule(rule, ruleId);
+    }
+
+    @Delete(':clusterId/rules/:ruleId')
+    async deleteFalcoRule(
```
jacobbeasley, line 226: Needs access controls I think

jacobbeasley: Admins only

dash/backend/src/modules/falco/controllers/falco.controller.ts
kyle-intelletive: Added in latest commit, now includes cluster ID, namespace, image, and rule name

dash/backend/src/migrations/2023062319298482489_add_falco_rules.ts
```diff
@@ -0,0 +1,23 @@
...
+        table.bigInteger('created_at')
+            .notNullable()
+            .defaultTo(knex.raw('getCurrentUnixTimestamp() * 1000'));
+        table.bigInteger('deleted_at');
+        table.string('action');
+        table.string('falco_rule');
+        table.string('namespace');
+        table.string('image');
+        knex.raw('create index idx_cluster_id on falco_rules(cluster_id)');
+        knex.raw('create index idx_cluster_id_created_at on falco_rules(cluster_id, created_at desc)');
```
jacobbeasley, line 17: Line 17 actually covers line 16 as well, so I think line 16 is unnecessary

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -253,4 +261,55 @@ export class FalcoService {
...
+        if (useCached) {
+            const cachedData = this.ruleCache.get(clusterId)?.data;
+            if (cachedData) {
+                return cachedData;
+            }
+        }
+
+        // If we don't use the cache, or there was no unexpired cache, get the rules from the DB and cache them.
+        const rules = await this.falcoDao.listActiveFalcoRulesForCluster(clusterId);
+        this.ruleCache.set(clusterId, DataCache.cacheFor(rules, 60000)); // 60000ms = 1min
```
jacobbeasley, line 279: nice

PR #125, opened 2023-06-27
Falco rules now accept regexes for the falco rule and image

kyle-intelletive:
- now accepts & validates that a Falco Rule's falco rule & image are regex bodies.
- Actually displays error message for failure to create, update or delete falco rule


dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -298,9 +298,9 @@ export class FalcoService {
                 const namespaceMatches = !rule.namespace
```
jacobbeasley, line 298: I wonder if we should support a list of namespaces?

PR #127, opened 2023-06-28
Use Alertservice to replace alert()

charisprose:


jacobbeasley:
@charisprose  Be sure to list in our ticket which pages were changed so we can regression test

PR #128, opened 2023-06-28
Fix for Gatekeeper sync issue 

kyle-intelletive:
# Main issue
the calculateExceptionBlock function in the exceptionBlockService builds the body of a rego array.
When their are no image-specific exceptions, it defaults to adding '{true, true}'. Unlike the similar logic for non-image specific exceptions, there is no comma after this {true, true}.

If an exception hits this case, and has at least one more exception processed after it, this will cause a rego syntax error.

Works by coincidence if this kind of exception was the only exception or the exception applied last, as then that comma would be optional. (having it for the last doesn't matter as trailing commas are legal in rego)

## Side issue
When getting the gatekeeper exceptions, exceptions were being filtered to only include ones whose issue_identifier was an empty string. However, new gatekeeper exceptions would have that as null rather than an empty string, and if someone used edit mode to repurpose an existing exception it could actually have something in it.

I simply removed this condition, as that field is effectively unused for this kind of exception.

# Misc other changes
- .env.sample GATEKEEPER_TEMPLATE_DIR was not set to the correct path, made match default from gatekepper.ts in config directory
- Fixed some linter complaints
  - if...else with no body to if converted to if
  - several let's changed to const's for variables that were never modified.

### DIFF
--- a/dash/backend/.env.sample
+++ b/dash/backend/.env.sample
@@ -109,4 +109,4 @@ IS_LOCAL=True
 #######################################################
 #./config/gatekeeper.ts
 #######################################################
-GATEKEEPER_TEMPLATE_DIR=gatekeeper-templates
+GATEKEEPER_TEMPLATE_DIR=dist/gatekeeper-templates
--- a/dash/backend/src/modules/command-line/services/exception-block.service.ts
+++ b/dash/backend/src/modules/command-line/services/exception-block.service.ts
@@ -252,19 +252,16 @@ export class ExceptionBlockService {
         for (const namespace of exception.namespaces) {
             namespaceRego += ` pod.metadata.namespace == "${namespace.name}",\n `;
         }
-        if (imagePattern === '') {
-
-        } else {
+        if (imagePattern !== '') {
             for (const namespace of exception.namespaces) {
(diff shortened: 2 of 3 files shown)


dash/backend/.env.sample
```diff
@@ -109,4 +109,4 @@ IS_LOCAL=True
 #######################################################
 #./config/gatekeeper.ts
 #######################################################
-GATEKEEPER_TEMPLATE_DIR=gatekeeper-templates
+GATEKEEPER_TEMPLATE_DIR=dist/gatekeeper-templates
```
jacobbeasley, line 112: Is this change just for local dev?

dash/backend/src/modules/command-line/services/exception-block.service.ts
```diff
@@ -252,19 +252,16 @@ export class ExceptionBlockService {
         for (const namespace of exception.namespaces) {
             namespaceRego += ` pod.metadata.namespace == "${namespace.name}",\n `;
         }
-        if (imagePattern === '') {
-
-        } else {
+        if (imagePattern !== '') {
```
jacobbeasley, line 255: haha nice

dash/backend/.env.sample
kyle-intelletive: Yes, this is .env.sample, which only exists to be copied and tweaked for people to create their own .env

PR #130, opened 2023-06-28
azure oauth now testable locally

kyle-intelletive:
Basically just made the oauth use the frontend url instead of the server url.

### DIFF
--- a/dash/backend/.env.sample
+++ b/dash/backend/.env.sample
@@ -5,6 +5,7 @@ SERVER_NAME=NestJs Server
 SERVER_HOST=localhost
 SERVER_PORT=3200
 SERVER_BASE_URL=http://mine-sweeper.ngrok.io
+SERVER_FRONTEND_URL=http://localhost:4200
 
 #######################################################
 #./config/security.ts
--- a/dash/backend/src/config/server.ts
+++ b/dash/backend/src/config/server.ts
@@ -3,5 +3,5 @@ export default {
   host: process.env.SERVER_HOST || '0.0.0.0',
   port: process.env.SERVER_PORT || 3000,
   baseUrl: process.env.SERVER_BASE_URL || 'http://localhost:3000/',
-  frontendUrl: process.env.SERVER_FRONTEND_URL || process.env.SERVER_BASE_URL
+  frontendUrl: process.env.SERVER_FRONTEND_URL || process.env.SERVER_BASE_URL || 'http://localhost:3000/'
 }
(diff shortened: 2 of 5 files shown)


dash/backend/.env.sample
```diff
@@ -5,6 +5,7 @@ SERVER_NAME=NestJs Server
 SERVER_HOST=localhost
 SERVER_PORT=3200
 SERVER_BASE_URL=http://mine-sweeper.ngrok.io
+SERVER_FRONTEND_URL=http://localhost:4200
```
jacobbeasley, line 8: How is this different than baseUrl?

kyle-intelletive: When running dash locally, you will be running the API on port 3200, and the UI on port 4200.

SERVER_BASE_URL should point to the API's host, and be used for things like Trivy that need to talk to the API.

SERVER_FRONTEND_URL should point to the UI, and is used for things like the OAuth redirect, since that needs to come back to the UI.

Most of the time when these are deployed, they will run on the same host (I.E dev-m9sweeper.intelletive.com), which is why the config service defaults the frontend url to the server url when it is not directly set.

jacobbeasley: Yeah, but won't the UI port (4200) reverse proxy to 3200 anyways for all API requests?

jshoberg: Does it need to point to the backend specifically? I thought it was set up to automatically redirect API requests from the frontend to the backend.

At least when running on my machine, I pointed the OAuth redirect to port 4200 and had no issues.

jacobbeasley: Or is there a case here where that won't happen?

jacobbeasley: I suppose we could see there could be cases where its deployed separately.

jacobbeasley: Lets document this in the /docs/ too and explain in the docs that the frontend url only needs to be set if the frontend and backend url is not the same, such as if they were fronting static docs on a different domain name using a web server like nginx or something.

jacobbeasley: So, base url is *supposed* to be whatever url that a user would go to in their browser.

jacobbeasley: Its used a couple contexts right now: 
 - password reset emails (and any other emails)
 - webhook code generation for kube bench, kube hunter, falco, etc

jacobbeasley: So I'm still not sure why we need this frontend url

kyle-intelletive: I added an extra sentence to its blurb in the docs.

Otherwise, if you would prefer I can abandon this PR, and we can create an item to get rid of the frontendUrl altogether.

jacobbeasley: Yeah, I'm thinking BASE_URL is probably sufficient.

PR #132, opened 2023-06-29
Dash - Add API Metrics Endpoint

becky-intelletive:


dash/backend/src/modules/metrics/services/prometheus-environment-metrics.service.ts
```diff
@@ -0,0 +1,150 @@
...
+      this.numMediumCVEs.labels(cluster.name).set(summary.totalMedium);
+      this.numFixableMediumCVEs.labels(cluster.name).set(summary.totalFixableMedium);
+
+      this.numLowCVEs.labels(cluster.name).set(summary.totalLow);
+      this.numFixableLowCVEs.labels(cluster.name).set(summary.totalFixableLow);
+
+      this.numNegligibleCVEs.labels(cluster.name).set(summary.totalNegligible);
+      this.numFixableNegligibleCVEs.labels(cluster.name).set(summary.totalFixableNegligible);
+      return;
+    } catch (e) {
```
jacobbeasley, line 146: Not sure if the return statement is needed

dash/backend/src/app.module.ts
```diff
@@ -141,9 +141,8 @@ const myFormatter = info => {
         enabled: true
       },
     }),
-    ScheduleModule.forRoot(),
```
jacobbeasley, line 144: Wait, did we intend to remove this? I think another change Kyle worked on might have required this.

dash/backend/src/modules/image/dao/image.dao.ts
```diff
@@ -178,12 +178,12 @@ export class ImageDao {
         });
     }
 
-    async countImage(searchClause: any): Promise<any> {
+    async countImage(searchClause: any): Promise<string> {  // this was always returning a string. now it's typed appropriately.
```
kyle-intelletive, line 181: Wouldn't it make more sense to cast this to a number rather than having a count variable be a string?

I checked, and this function is hooked up to an endpoint, although the frontend does not call that endpoint, it does expect a number. I'm guessing that this worked by coincidence because the string was just being passed to a paginator which would cast the type autoamtically.

dash/backend/src/modules/image/services/image.service.ts
```diff
@@ -28,7 +29,7 @@ export class ImageService {
     ) {
     }
 
-    async getAllImagesByClusterId(clusterId: number): Promise<{total: number, listOfImages: ListOfImagesDto[]}> {
+    async getAllImagesByClusterId(clusterId: number): Promise<{total: string, listOfImages: ListOfImagesDto[]}> {
```
kyle-intelletive, line 32: Once again, this should probably have the total as a number. Based on a quick check, it would probably make sense to have the DAO cast the count to a number so that it is accurate for everywhere.

With current usages of this it will not make a difference, but if any math is attempted on this total, unexpected behavior is likely with it as a string

dash/backend/src/modules/reports/dao/reports.dao.ts
```diff
@@ -254,13 +265,37 @@ export class ReportsDao {
             query.limit(options.limit);
         }
 
-        const queryResults = await query.then((response) => {
-            return plainToClass(ReportsRunningVulnerabilitiesDto, response);
+        const queryResults: ReportsRunningVulnerabilitiesDto[] = await query.then((response) => {
+          // plainToInstance returns a single even when it's an array --> proper typing requires longer form
```
kyle-intelletive, line 269: This comment is inaccurate, and you shouldn't need to be doing this mapping here. The functions js doc is: 'Converts plain (literal) object to class (constructor) object. Also works with arrays'.

plainToInstance(ReportsRunningVulnerabilitiesDto, response) will return a single object if response is a single object, or process & return the entire array if response is an array.

If your IDE isn't liking this, then it is likely configured with an incorrect linteror typescript config, or you need to give response a type since right now it is defaulting to any, it should probably be any[])

dash/backend/src/app.module.ts
becky-intelletive: Yes. It was only added to account for the old metrics cron job.

dash/backend/src/modules/image/dao/image.dao.ts
becky-intelletive: It would make sense, but I didn't want to change the functionality. Like the comment indicates, it was always returning a string and not a number, so I didn't want to change the functional type.

kyle-intelletive: This function was not being used before the changes you made, so it would be the perfect opportunity to fix the typing.
Why make a comment explaining that the code is weird rather than make it better when there are no regressions?

becky-intelletive: Because typically you don't want to change what's functionally being returned via an API just in case someone or something is depending on it being odd.
I can change it though.

PR #133, opened 2023-06-29
Chart - Add Standard Metrics

becky-intelletive:


manifests/charts/dash/values.yaml
```diff
@@ -43,6 +43,38 @@ service:
 
 prometheusOperator:
   enabled: true
```
jacobbeasley, line 45: Should we be defaulting this to true? If they don't have kube-prometheus-operator installed are we going to end up deploying anything that will cause m9sweeper not to install?

jacobbeasley: @jason-woodman @jasonWoodman  Probably a question for you ^^^

jacobbeasley:
We probably should update docs, too, to include details about how to enable this if they are using kube-prometheus operator and a couple sample screenshots of the dashboards.

manifests/charts/dash/templates/prometheus-rules-default.yaml
```diff
@@ -0,0 +1,42 @@
...
+          annotations:
+            description: Fires when the number of new major + critical CVEs in the last 10m exceeds 20
+            runbook_url: url_here
+            summary: The number of new CVEs breached 20
+          expr: (delta(num_critical_cves[10m])+delta(num_major_cves[10m])) > 20
+          labels:
+            severity: warning
+    {{- end }}# closes if .Values.prometheusOperator.defaultRules.newCVEs.enabled
+    {{- if .Values.prometheusOperator.defaultRules.ruleTwo.enabled }}
+    - name: ruleTwo
```
kyle-intelletive, line 30: This looks kind of a like a placeholder or unfinished.

manifests/charts/dash/values.yaml
jacobbeasley: Becky is going to set it to enabled:false and then we can merge.

PR #135, opened 2023-06-29
use logging service instead of console.log

becky-intelletive:


dash/backend/src/modules/shared/services/mine-logger.service.ts
```diff
@@ -21,42 +21,52 @@ export class MineLoggerService extends ConsoleLogger {
...
+          }, MineLoggerService.name);
+        }).catch(e => {
+          this.log({
+            label: 'Failed to send error notification email',
+            data: { mailSendResponse: e.message }
+          }, MineLoggerService.name);
+        });
+      }
+    } catch (e) {
+      console.log('An error occurred while attempting to log an error message. Information: ', message, trace, context, { loggingError: e });
```
jacobbeasley, line 69: Why console.log here?

dash/backend/src/main.ts
```diff
@@ -208,15 +197,9 @@ async function bootstrap() {
   try {
     await m9sCronService.updateExceptionAndImageMetrics();
   } catch (e) {
-    console.log('There was an error setting the initial v1 exception and image metrics');
+    logger.log('There was an error setting the initial v1 exception and image metrics', e, 'bootstrap');
```
kyle-intelletive, line 200: This should probably a logger.error

dash/backend/src/modules/auth/services/ldap/ldap-auth-provider.ts
```diff
@@ -1,10 +1,15 @@
 import { AuthConfiguration } from '../../models/auth-configuration';
 import {ILDAP} from '../interfaces/ILDAP';
+import { MineLoggerService } from '../../../shared/services/mine-logger.service';
 
 export abstract class LdapAuthProvider implements ILDAP {
 
   _authConfiguration: AuthConfiguration;
 
+  constructor(
+    protected logger: MineLoggerService,
```
kyle-intelletive, line 10: Is there a reason to inject this if it isn't being used?

dash/backend/src/modules/shared/services/mine-logger.service.ts
becky-intelletive: Because this is in the logging service. If all else fails and it can't log the official way, then there'll at least be something present.

jacobbeasley: haha oh right :D

PR #139, opened 2023-07-03
Global falco rule implementation

kyle-intelletive:
- Falco rule section removed from cluster specific cluster settings
- Falco rule DB refactor
   - Now uses join table for namespace & cluster
   - Has field to designate if applicable to all namespace & cluster
   - original cluster & namespace field removed
- Falco Settings page added to organization settings page
   - Displays boxes with all existing Falco rules as before
   - Add/Edit/Delete available here (for admins only)
   - Add/Edit Modal adjusted to account for multiple namespaces/clusters
- Validation Updated
   - Accounts for changes to allow multiple clusters/namespaces


dash/backend/src/migrations/2023062816298482489_falco_rule_join_tables.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<any> {
+    await knex.schema.alterTable('falco_rules', (table) => {
```
jacobbeasley, line 4: We need to move existing data into the new tables, so maybe it needs to be: 
 - create new tables
 - copy old data into new tables (insert into ... select *)
 - then drop the old columns

jacobbeasley: The other alternative would just be to use an array[] data type or the cluster_ids fields

dash/backend/src/modules/falco/controllers/falco.controller.ts
```diff
@@ -199,48 +186,57 @@ export class FalcoController {
...
         return await this.falcoService.createFalcoRule(rule);
     }
 
-    @Get(':clusterId/rules')    @AllowedAuthorityLevels( Authority.SUPER_ADMIN, Authority.ADMIN, Authority.READ_ONLY )
+    @Get('/rules')    @AllowedAuthorityLevels( Authority.SUPER_ADMIN, Authority.ADMIN, Authority.READ_ONLY )
     @UseGuards(AuthGuard, AuthorityGuard)
-    async listActiveFalcoRulesForCluster(@Param('clusterId') clusterId: number): Promise<FalcoRuleDto[]> {
-        return this.falcoService.listActiveFalcoRulesForCluster(clusterId);
-
+    async listActiveFalcoRulesForCluster(): Promise<FalcoRuleDto[]> {
```
jacobbeasley, line 202: Are you filtering on the frontend now?

dash/backend/src/modules/falco/dao/falco.dao.ts
```diff
@@ -368,30 +368,147 @@ export class FalcoDao {
...
+            rule.allClusters = !rule?.clusters?.length;
+            let rawNamespaces = [];
+            if (!rule.allNamespaces) {
+                rawNamespaces = rule.namespaces.map(ns => {
+                    return { namespace: ns }
+                })
+            }
+            let rawClusters = [];
+            if (!rule.allClusters) {
+                rawClusters = rule.clusters.map(id => {
```
jacobbeasley, line 384: id => ({ cluster_id: id })

dash/backend/src/modules/falco/service/falco.service.ts
```diff
@@ -275,28 +284,29 @@ export class FalcoService {
...
-        return this.falcoDao.updateFalcoRule(rule, ruleId);
+    async updateFalcoRule(rule: FalcoRuleCreateDto, ruleId: number): Promise<FalcoRuleDto> {
+        return this.falcoDao.updateFalcoRule(rule, ruleId)
+          .then(() => this.getFalcoRuleById(ruleId));
     }
 
-    async deleteFalcoRule(clusterId: number, ruleId: number): Promise<FalcoRuleDto> {
-        const edits = Object.assign(new FalcoRuleDto(), { deletedAt: Date.now() });
-        return this.falcoDao.updateFalcoRule(edits, ruleId);
+    async deleteFalcoRule(clusterId: number, ruleId: number): Promise<number> {
```
jacobbeasley, line 297: If we're not using clusterId, should we remove it?

dash/frontend/src/app/core/services/custom-validator.service.ts
```diff
@@ -31,11 +31,13 @@ export class CustomValidatorService {
...
   }
 
-  /** Group level validator. Requires the at least one of the controls with the specified keys to have a value */
-  atLeastOne(keys: string[]): ValidatorFn {
+  /** Group level validator. Requires the at least one of the controls with the specified keys to have meet its criteria */
+  atLeastOne(params: { key: string, condition: (val: any) => boolean}[]): ValidatorFn {
     return (group: FormGroup) => {
       const controls = group?.controls;
-      const exists = keys.some(key => !!controls?.[key]?.value);
+      const exists = params.some(param => {
```
jacobbeasley, line 38: params.some(params => param.condition(controls?.[param.key]?.value)

dash/backend/src/migrations/2023062816298482489_falco_rule_join_tables.ts
kyle-intelletive: I didn't think it was worth doing this.

kyle-intelletive: (As the only DB with data in this table is my local copy)

dash/backend/src/modules/falco/controllers/falco.controller.ts
kyle-intelletive: There is currently no filtering, as there is now only the list of all rules

dash/backend/src/modules/falco/dao/falco.dao.ts
kyle-intelletive: Latest commit uses this syntax for all 4 maps in create & update dao functions

PR #175, opened 2023-07-19
More UI fixes

charisprose:


dash/frontend/src/app/modules/private/pages/change-password/change-password.component.html
```diff
@@ -3,7 +3,7 @@
     <mat-card appearance="outlined" class="change-password-mat-card">
       <mat-card-header>
         <mat-card-title>
-          <h3>Change Password</h3>
+          <a>Change Password</a>
```
becky-intelletive, line 6: Is an <a> tag the right element for this?

dash/frontend/src/app/modules/private/private.component.html
```diff
@@ -30,31 +30,38 @@
     </div>
   </mat-toolbar-row>
   <mat-menu #menu="matMenu">
-    <button mat-menu-item>
+    &nbsp;
```
becky-intelletive, line 33: There are lots of `&nbsp;`s added to this file - can we remove them?

dash/frontend/src/app/modules/private/private.component.ts
```diff
@@ -217,7 +217,6 @@ export class PrivateComponent implements OnInit, AfterViewInit, OnDestroy {
       data: {}
     });
     confirmDialog.afterClosed().subscribe(result => {
-      this.directive.ngOnInit();
```
becky-intelletive, line 220: What purpose did this serve? I'm wondering if this might not be required somewhere.
Thoughts @kyle-intelletive?

dash/frontend/src/app/modules/public/pages/login/login.component.html
```diff
@@ -35,7 +35,7 @@
             <input id="password" matInput type="password" formControlName="password">
           </mat-form-field>
           <div class="form-action-button" *ngIf="selectedInSiteCredentialAuthenticationMethod?.type === 'LOCAL_AUTH' || selectedInSiteCredentialAuthenticationMethod?.type === 'LDAP'">
-            <button id="submit'" mat-raised-button type="submit">Login</button>
+            <button id="submit'" mat-raised-button mat-button type="submit">Login</button>
```
becky-intelletive, line 38: Can we just make this a <mat-button> instead of a <button>?

becky-intelletive: And we can probably set color="primary" here

dash/frontend/src/app/modules/public/pages/login/login.component.scss
```diff
@@ -32,6 +33,7 @@ div {
                 margin: 5px 0;
                 button{
                   width: 100%;
+                  background-color:  mat.get-color-from-palette(mat.$gray-palette, 300);
```
becky-intelletive, line 36: This is likely unnecessary (see comment in the html file)

dash/frontend/src/app/modules/private/private.component.ts
kyle-intelletive: I can take a look

kyle-intelletive: It looks like directive is a ViewChild reference to a UserProfileImageDirective. My assumption is that this code used to cause that to reload the user profile picture.

Since we don't atually allow that, we probably don't need anything related to that ViewChild, or this subscription to the dialogref.afterclosed()

dash/frontend/src/app/modules/public/pages/login/login.component.html
kyle-intelletive: Material design doesn't like having mat-raised-button and mat-button on the same element, you likely just want to set color="primary" as Becky suggested, and remove the plain mat-button directive.

charisprose: It used to be grey I thought. If blue works, that's fine for me.

dash/frontend/src/app/modules/public/pages/login/login.component.scss
charisprose: This will not be necessary if we choose to use blue (primary) button.

dash/frontend/src/app/modules/private/private.component.html
charisprose: This is for the extra space requested for the top-right-dropdown menu.

dash/frontend/src/app/modules/private/pages/change-password/change-password.component.html
becky-intelletive: The document refers to the link in the menu. The menu's "Change password" should be a link.
image

dash/frontend/src/app/modules/private/private.component.html
becky-intelletive: The extra space is a problem, not the end goal ;)

dash/frontend/src/app/modules/private/private.component.ts
becky-intelletive: In that case, should we add that to the list of this to tear out?

dash/frontend/src/app/modules/public/pages/login/login.component.html
becky-intelletive: Yes, we want it to be the primary color for consistency across the platform :)

dash/frontend/src/app/modules/private/pages/change-password/change-password.component.html
charisprose: Change Password is already a link to /private/change-password.

dash/frontend/src/app/modules/private/private.component.ts
kyle-intelletive: Yes, while the directive is still actually being used, as far as I can tell it is entirely useless, since all it has is a constructor with nothing in it

dash/frontend/src/app/modules/private/pages/change-password/change-password.component.html
becky-intelletive: It isn't - it's a button. It redirects to /private/change-password, but it isn't a link.

dash/frontend/src/app/modules/private/private.component.ts
charisprose: @kyle-intelletive After removing that line of code, name change was updated. Do you want me to leave this as is for now?

kyle-intelletive: Yeah don't worry about it, if we remove the directive we can remove the rest at the point.

dash/frontend/src/app/modules/private/pages/change-password/change-password.component.html
charisprose: @becky-intelletive The items on this dropdown menu are all buttons. Help me to understand the thinking behind changing this from a button to a link.

becky-intelletive: It's a link, so it should act like one.
You can make it a button/link situation like the documentation maybe? Would that make more sense?

becky-intelletive (approved):

PR #176, opened 2023-07-19
cluster info page formatting

becky-intelletive:


dash/frontend/src/app/modules/private/pages/cluster/cluster-info/cluster-info.component.ts
```diff
@@ -68,24 +86,15 @@ export class ClusterInfoComponent implements OnInit {
...
-    if (val !== this.width) {
-      this.width = val;
-    }
-  }
-
-  get scrWidth(): number {
-    return this.width;
+  ngAfterViewInit() {
+    this.isHandsetOrXS$.subscribe((newIsHandsetOrXS) => {
+      this.isMobileDevice = newIsHandsetOrXS;
```
kyle-intelletive, line 91: you probably want a .pipe(takeUntil(this.unsubscribe$) here.

Also, since it looks like this is the only place you are using the isHandsetOrXS$, you probably don't need the shareReplay() (or to have it as a class property)

becky-intelletive: @kyle-intelletive - updated. Could you take another look?

kyle-intelletive (approved):

PR #188, opened 2023-07-24
Update gatekeeper dialogues and recommendation text behavior

jshoberg:


becky-intelletive (dismissed):

becky-intelletive:
I'll let @kyle-intelletive review as well to check the subscription changes, but it looks good to me!

dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.ts
```diff
@@ -36,20 +38,35 @@ export class GateKeeperComponent implements OnInit {
 
   ngOnInit(): void {
     this.checkGatekeeperStatus();
-    this.gatekeeperConstraintTemplates$ = this.gateKeeperService.getGateKeeperConstraintTemplatesByCluster(this.clusterId);
+    this.gateKeeperService.getGateKeeperConstraintTemplatesByCluster(this.clusterId)
+      .pipe(takeUntil(this.unsubscribe$))
```
kyle-intelletive, line 42: This looks like a call to an API. Assuming that is the case, this should be a take(1) rather than a takeUntil, and in that case I think you could remove the unsusbscribe$ creation and desctruction you added/

becky-intelletive (approved):

PR #190, opened 2023-07-25
Report page layout updates

jshoberg:


dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
```diff
@@ -86,73 +88,75 @@
           <div class="page-table">
             <mat-table [dataSource]="dataSource">
               <ng-container matColumnDef="image" class="column-word-break">
-                <mat-header-cell *matHeaderCellDef>Image</mat-header-cell>
-                <mat-cell *matCellDef="let vulnerability">{{vulnerability.image}}</mat-cell>
+                <mat-header-cell class="min-width-8" *matHeaderCellDef>Image</mat-header-cell>
+                <mat-cell class="min-width-8" *matCellDef="let vulnerability">{{vulnerability.image}}</mat-cell>
```
kyle-intelletive, line 92: Does this work when the data in the rows is larger than the title, and larger than that min-width size?

Elsewhere we've been setting the min-width on the generated class name for the column, so I'm not sure if setting it directly on the cells will behave differently.

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
```diff
@@ -70,14 +70,16 @@
       <mat-card appearance="outlined">
```
becky-intelletive, line 70: Remove `appearance="outlined"` if you don't mind :)

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
jshoberg: It seemed to when I was testing it, the data breaks into a new line if necessary with the row heights resizing. I'm not sure if that's the behavior you expect on these tables.

If not I can look at defining it on generated class names.

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
becky-intelletive: Also if the table should be scrollable, we have classes for that.

becky-intelletive (changes requested):

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
kyle-intelletive: My concern was that the cells in the same columns might end up with different widths and mess up the alignment of the table. So if that's not the case it is good.

becky-intelletive (approved):

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
becky-intelletive: If it's scrolling properly on a small screen, that's good enough for me.

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
jshoberg: Alright, yeah I haven't seen any instance where the columns have fallen out of alignment when testing locally. If something crops up during testing I can circle back to it, but so far it's looked good.

dash/frontend/src/app/modules/private/pages/reports/running-vulnerabilities/running-vulnerabilities.component.html
becky-intelletive: You should double check to be sure.
Here I have a <mat-card-content> with a nested <mat-table>:
image

becky-intelletive: When it's nested like:
`        <mat-card class="scrollable-table-card">`
`          <mat-card-content>`
`            <div class="table-wrapper">`
`              <mat-table [dataSource]="dataSource" id="falco-table">`
and it has scss setting the min-widths of each column,
it works

jshoberg: The tables scroll fine on small screens locally.

image

If I remember right the issue was the card itself had a min width set at 800px, removing that and it sizes down as expected and mat-card-content handles the table scrolling.

PR #198, opened 2023-07-27
User dropdown fixes

jshoberg:


dash/frontend/src/app/modules/private/private.component.scss
```diff
@@ -88,3 +88,7 @@ mat-toolbar.fixed {
     }
   }
 }
+
+a:visited {
+  color: blue;
```
becky-intelletive, line 93: We _don't_ want it to change color after it's been visited lol

becky-intelletive (changes requested):

dash/frontend/src/app/modules/private/private.component.scss
jshoberg: That's what it's doing - by default hyperlinks are blue, with the visited attribute setting them to purple. This is just overriding that behavior to keep the color the same.

jshoberg: If that's the case, just pushed an update to unify the text color

becky-intelletive (approved):

PR #199, opened 2023-07-27
User List cleanup

becky-intelletive:
- Add phone number mask directive for inputs
- Fix sorting for list of users
- Clean up scanner list on create/edit policy page
- Clean up Settings pages header

### DIFF
--- a/dash/backend/src/modules/user/dao/user.dao.ts
+++ b/dash/backend/src/modules/user/dao/user.dao.ts
@@ -81,42 +81,59 @@ export class UserDao {
                         .andWhere( 'u.id', '=', userId)
                         .groupByRaw('al.entity_type, al.entity_id')
                         .returning('fail_count');
-                       
+
     return queryResult[0].fail_count;
   }
 
-  async loadUser(searchClause?: any, page = 0, limit = 10,
-                 sort: {field: string; direction: string; } = {field: 'id', direction: 'asc'}):
-      Promise<UserProfileDto[]> {
+  async loadUser(
+    searchClause?: any,
+    page = 0,
+    limit = 10,
+    sort: {field: string; direction: string; } = {field: 'id', direction: 'asc'}
+  ): Promise<UserProfileDto[]> {
     const knex = await this.databaseService.getConnection();
+    const processedSort = structuredClone(sort);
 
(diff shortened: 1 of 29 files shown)


dash/backend/src/modules/user/dao/user.dao.ts
```diff
@@ -81,51 +81,70 @@ export class UserDao {
...
            'authority_levels.id as _authorities__id', 'authority_levels.name as _authorities__name'
          ])
       .leftJoin('user_authorities as user_auth', function () {
         this.on('u.id', '=', 'user_auth.user_id').andOn('user_auth.active', '=', knex.raw('?', [true]));
       }).leftJoin('authority_levels', function () {
         this.on('user_auth.authority_id', '=', 'authority_levels.id');
-      });
+      }).orderBy(processedSort.field, processedSort.direction);
 
+    const sqlQuery = query.toSQL();
```
kyle-intelletive, line 146: I assume these 2 variables are residual from debugging and should be cleaned up unless I'm missing something

becky-intelletive: Cleaned

dash/frontend/src/app/modules/shared/directives/phone-mask.directive.ts
```diff
@@ -0,0 +1,65 @@
+import { Directive, Input, OnInit, OnDestroy } from '@angular/core';
```
kyle-intelletive, line 1: This looks like it only supports US phone numbers, I believe the simpler validation we had before was so that we could support international phone numbers without pulling in a "real" phone validation library. May want to double check with Jake/Jason before maknig this change.

becky-intelletive: There was no validation before. I can pull it into a separate PR though.

kyle-intelletive (dismissed):

dash/frontend/src/app/modules/shared/directives/phone-mask.directive.ts
becky-intelletive: Recheck

kyle-intelletive (approved):

PR #216, opened 2023-08-01
Frontend: npm yaml update

becky-intelletive:


brandan-schmitz (approved):

brandan-schmitz:
Was there a YAML package in the backend as well? I cannot remember. If so maybe we want to update it there too so they are both the same.

kyle-intelletive (approved):

PR #223, opened 2023-08-02
Remove tableify and rimraf

charisprose:


dash/backend/src/modules/auth/controllers/non-redirectable-login.controller.ts
```diff
@@ -103,6 +104,14 @@ export class NonRedirectableLoginController {
     async localAuthLoginAction(@Body() loginDto: LoginDto): Promise<any> {
         this.prometheusV1Service.loginCounter.inc(1);
         const users: UserProfileDto[] = await this.userProfileService.loadUserByEmail(loginDto.userName);
+        // array
+        const sampleJson = ['banana', 'apple', 'peach'];
+        // message object const sampleJson = {label: 'Error creating webhook for cluster', data: { clusterId: 5 }};
+        // null const sampleJson = null;
+        // text with symbols const sampleJson = '09:36:29.597260967: Notice Unexpected connection to K8s API Server from container (command=python -u /app/sidecar.py pid=46515 k8s.ns=monitoring-system k8s.pod=prometheus-stack-grafana-7594dc6d95-cjfjj container=87d81d268f9c image=quay.io/kiwigrid/k8s-sidecar:1.22.0 connection=10.244.2.25:35680->10.0.0.1:443)';
+        console.log(sampleJson);
```
jacobbeasley, line 112: Is this lefta round by accident?

dash/backend/src/util/tablify.ts
```diff
@@ -0,0 +1,43 @@
+export function tablify(obj: any, depth = 0, maxdepth = 10) {
```
becky-intelletive, line 1: Where is this function from?
Wondering if we should c+p the Tablify class from the frontend. I did a bit of bug fixing on it :)

becky-intelletive (changes requested):

dash/backend/src/util/tablify.ts
jacobbeasley: It looks a bit similar to this - https://github.com/m9sweeper/m9sweeper/blob/974365e590265f452b62c1b00facf1b94820d79a/dash/frontend/src/app/core/pipe/tablify.pipe.ts#L6

jacobbeasley: Though I don't think we want the backend to import PipeTransform, so we wouldn't copy paste the whole thing - just parts of it (maybe that's what Charis did). @becky-intelletive  Do you see a specific change required here?

becky-intelletive: You're right. On closer inspection it looks like it was a c+p :)

dash/backend/src/modules/auth/controllers/non-redirectable-login.controller.ts
charisprose: I gave it a test after I added the function. Might have missed to remove them. The lines will be removed.

dash/backend/src/util/tablify.ts
charisprose: @jacobbeasley The frontend pipe was from Angular. Per the instruction from the relevant ticket, this was modified a bit to provide the same usage. Let me know if you want any further changes.

becky-intelletive (dismissed):

becky-intelletive (approved):

dash/backend/src/email-templates/error-email.hbs
```diff
@@ -30,7 +30,7 @@
         </p>
 
         <p>
-            {{{data}}}
+            {{data}}
```
kyle-intelletive, line 33: Did you change from using something besides handlebars?

Handlebars does use triple brackets

becky-intelletive: Good catch!

becky-intelletive (approved):

PR #234, opened 2023-08-04
Cluster setup dialog fixes

jshoberg:


dash/frontend/src/app/modules/private/pages/cluster/add-cluster-wizard/add-cluster-wizard.component.html
```diff
@@ -76,10 +76,10 @@ <h3>
 
       <div class="row">
         <div class="col-xs-12 col-sm-6 center-xs start-sm display-flex mb-half-rem">
-          <button mat-raised-button color="accent" matStepperPrevious>Prev Step</button>
+          <button [disabled]="true" mat-raised-button color="accent" matStepperPrevious>Prev Step</button>
```
kyle-intelletive, line 79: Is it intentional this button is being permanently disabled?

If so, we could probably remove it, or replace it with a close dialog button

jshoberg: I think it was just there to have a consistent layout with previous/next buttons in the dialogue, I think it look clearer without the unnecessary button though.

kyle-intelletive (approved):

PR #239, opened 2023-08-07
More Cleanup

becky-intelletive:
- clean up "running vulnerabilities" table
- removed some unnecessary classes in styles.scss
- clean up image scan result page's table
- pods in namespace: make page fit content


dash/frontend/src/app/core/enum/CVETypes.ts
```diff
@@ -0,0 +1,16 @@
+
+export enum CVETypes {
```
kyle-intelletive, line 2: We have an enum called VaulnerabilitySeverity that has the same content, but with the values capitalized. Is this doing something different enough to have both?

becky-intelletive: Refactored to use that enum; put the abbreviations in that file too

kyle-intelletive (approved):

PR #259, opened 2023-08-15
Fix Trawler scanning

kyle-intelletive:
Changes made:
When the image digest is available, Trawler will use it to ensure the proper image is pulled. If it is not available (or is TMP_xxx) it will pull it by name and function as it previously did.

Image Id has changed... error can no longer happen, as Trawler will now pull the old version.

For exampl: If we have an nginx:latest running in our cluster that is actually nginx:1.20, when scanned, it will pull using the digest, so it will actually scan nginx:1.20 and give the expected result for that scan. With the old behavior, Trawler would pull nginx:latest by name, but would actually get nginx:1.25, and this would result in the Image ID does not match.... error, and no possibility of scanning that older nginx:latest image.

### DIFF
--- a/trawler/src/main/java/io/m9sweeper/trawler/ScanRunner.java
+++ b/trawler/src/main/java/io/m9sweeper/trawler/ScanRunner.java
@@ -46,7 +46,6 @@ public class ScanRunner {
         ArrayList<ScanResult> scanResults = new ArrayList<>(0);
 
         String newImageHash = "";
-        boolean errorEncountered = false;
 
         if (policies != null && policies.size() > 0) {
             for (PolicyWithScannerDto policyWithScannerDto: policies) {
@@ -87,13 +86,11 @@ public class ScanRunner {
                             scanResultBuilder.withSummary("");
                         } catch (Exception e) {
                             scanResultBuilder.withEncounteredError(true);
-                            errorEncountered = true;
                             scanResultBuilder.withSummary(e.getMessage());
                         }
                     }
                 } else {
                     scanResultBuilder.withEncounteredError(true);
-                    errorEncountered = true;
                     scanResultBuilder.withSummary("No Scanner found");
                     if (TrawlerConfiguration.getInstance().getDebug()) System.out.println("policyWithScannerDto.getScanners().size(): " + policyWithScannerDto.getScanners().size());
                 }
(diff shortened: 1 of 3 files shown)


trawler/src/main/java/io/m9sweeper/trawler/ScanRunner.java
```diff
@@ -183,21 +195,30 @@ public void scan() throws Exception {
...
         TrawlerScanResults body = new TrawlerScanResults();
         body.setData(imageTrawlerResultDtos);
         if (TrawlerConfiguration.getInstance().getDebug()) {
             System.out.println("Saving Scan Results: " + body.toString());
         }
         ImageScanResultSaveResponse complianceResponse = apiInstance.imageControllerSaveImageScanResults(body, new BigDecimal(message.getCluster().getId()), new BigDecimal(message.getImage().getId()));
-        if (complianceResponse.getData() != null && !complianceResponse.getData().isComplaint()) {
-            throw new NoncompliantException("Image Is not compliant", imageTrawlerResultDtos);
-        }
+        return complianceResponse.getData() == null || complianceResponse.getData().isComplaint();
```
kyle-intelletive, line 222: Now made this return a boolean, and use the boolean in the main body to throw this error. This prevents the error from short-circuiting and not returning both scan results when needed.

becky-intelletive (dismissed):

trawler/src/main/java/io/m9sweeper/trawler/ScanRunner.java
```diff
@@ -165,12 +168,21 @@ public void scan() throws Exception {
...
-                errorSummary = errorSummary.concat(". Note, using the latest tag for images is not recommended and can cause undesirable behavior");
+        // The hash has changed if the hash from the scan does not match the one from dash AND it is not a placeholder
+        boolean hashHasChanged = !scanConfig.getImage().getHash().equals(newImageHash) && (scanConfig.getImage().getHash() != null && !scanConfig.getImage().getHash().startsWith("TMP_"));
+        boolean compliant = true;
+
+        // If we have an error, or the hash has changed,
+        // Return an error response under the original hash
+        if (hashHasChanged || errorEncountered) {
+            String errorSummary = "";
+            if (hashHasChanged && !errorEncountered) {
```
jacobbeasley, line 179: Might be able to simplify logic to...

if (errorEncountered) {
errorSummary = imageTrawlerResultDtos.get(0).getSummary();
} else if (hashHasChanged) {
errorSummary = "Image ID does not match most recent image pulled from Trawler";
                if (scanConfig.getImage().getTag().contains("latest")) {
                    errorSummary = errorSummary.concat(". Note, using the latest tag for images is not recommended and can cause undesirable behavior");
                }
}

I think

jacobbeasley: Though I think what you have will work

kyle-intelletive: The if block that chooses the error Summary message is now flipped.

kyle-intelletive:
Using this branch/Pr for the changes to make Trawler pull with image digests when scanning with Trivy

jacobbeasley:
Woooooo

trawler/src/main/java/io/m9sweeper/trawler/framework/docker/DockerImage.java
```diff
@@ -57,6 +57,38 @@ public void setName(String name) {
...
+    public String buildFullPath(boolean includeRegistry, boolean includeHash) {
+        StringBuilder builder = new StringBuilder(32);
+        if (includeRegistry) {
+            builder.append(this.registry.getHostname())
+                    .append('/');
+        }
+
+        builder.append(this.name)
+                .append(":")
+                .append(this.tag);
```
kyle-intelletive, line 75: The tag is ignored when a SHA is present, so I left it as always being part of the string.

becky-intelletive (approved):

PR #271, opened 2023-08-22
Ensure dash imagePullPolicy is respected

brandan-schmitz:
Update the dash chart to ensure that the pullPolicy for the dash and kubesec images is actually respected and used. This is needed when installing in minikube and images have been manually loaded via the `minikube image load` command since you need to set the imagePullPolicy to `Never` in that situation.

### DIFF
--- a/manifests/charts/dash/templates/cronjob-exception-status-update.yaml
+++ b/manifests/charts/dash/templates/cronjob-exception-status-update.yaml
@@ -34,6 +34,7 @@ spec:
           containers:
             - name: {{ include "dash.fullname" . }}-scrape-history
               image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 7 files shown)


manifests/charts/dash/templates/cronjob-exception-status-update.yaml
```diff
@@ -34,6 +34,7 @@ spec:
           containers:
             - name: {{ include "dash.fullname" . }}-scrape-history
               image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
```
jacobbeasley, line 37: good idea

jacobbeasley (approved):

jacobbeasley:
Excellent, per the usual!

PR #274, opened 2023-08-24
Fixes for: falco sort, logs "floating" cluster compliance graph 

kyle-intelletive:
## Falco Logs (Ticket 6881)
The switch for which sort to use was wrapped in an if statement requiring specific sorts. Since there is no default sort, this was effectively preventing it from reaching the default sort case.

Changed it so it always uses the switch statement, and now defaults to date instead of id for default sort.
Also made UI start with date descending selected as the default sort to make this explicit

## Floating Cluster Logs (Ticket 6842)
Changed it from a max-height to a height for its container, so it is now sticky and always takes up the same amount of space on the bottom of the screen.

Also added a loader to the logs to make the infinite scroll more obvious.

## Cluster Compliance Graph (Ticket 6824)
2 changes made to solve for dates with no scans looking unscanned:
1. The graph now only includes dates where the cluster/cluster group existed.
     a. Effectively this means instead of always starting 28 days ago, it now starts 28 days ago, or when the cluster [group] was created - whichever was more recent
2. If there was no data for a day now defaults to 100% compliance instead of 0% compliance

### Other compliance graph bug
The cluster compliance graph on the cluster group page actually shows compliance over EVERY cluster and cluster group, rather than just its group. Added functionality so the compliance endpoint can work by cluster or cluster group.

### DIFF
--- a/dash/frontend/src/app/modules/private/pages/cluster/cluster-summary/cluster-summary.component.ts
+++ b/dash/frontend/src/app/modules/private/pages/cluster/cluster-summary/cluster-summary.component.ts
@@ -185,7 +185,7 @@ export class ClusterSummaryComponent implements OnInit, AfterViewInit, OnDestroy
   }
 
   getPodComplianceSummaryForAllClusters(clusterId: number) {
-    this.podService.getPodsComplianceSummary(clusterId)
+    this.podService.getPodsComplianceSummary({ clusterId })
         .pipe(take(1))
         .subscribe(response => {
       const complianceSummaryData = response.data;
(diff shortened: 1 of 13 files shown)


dash/frontend/src/app/modules/private/pages/cluster/cluster-summary/cluster-summary.component.ts
```diff
@@ -96,21 +96,23 @@ export class ClusterSummaryComponent implements OnInit, AfterViewInit, OnDestroy
   page = 0;
   limit = 10;
   clusterEvents: IClusterEvent[];
-  expandStatus: boolean;
-  expandSubscription: Subscription;
   resizeTimeout;
+  logLoaderName = 'cluster-logs-loader';
+  logsLoaderConfig: NgxUiLoaderConfig = {
+    bgsType: SPINNER.rectangleBounce,
+    bgsPosition: POSITION.centerCenter,
```
becky-intelletive, line 103: Were you able to validate that this positions it where you want it? (it looks like it'll put it in the center of the screen)

kyle-intelletive: Yes, the position is relative to its parent container, not the page. Here is a screenshot of where it is. (It is animated See rectangle-bounce here if you want to see that: https://tdev.app/ngx-ui-loader/demo/spinners)


Screenshot 2023-08-24 at 4 52 36 PM

becky-intelletive: Great!
I found with that loader that it often positioned it absolutely and it could be a pain to get it to stop.

becky-intelletive (approved):

PR #281, opened 2023-09-07
Mobile tweaks 2

kyle-intelletive:
# Misc Changes

## Created app-ellipsis-text component
This was used in several spaces to make things work better for smaller screens.
- Will display 2 lines of the text passed in, and then cut off with an ellipsis
- When clicked, will open the full text in a dialog, (optionally with a header labellng it)
  - this behavior can be disabled, in which case it will only display the text. 

## Added text-align-[size]-[direction] classes
Generated via the generic styles Becky added. Was useful for some responsive functionality

# Mobile fixes
- The gatekeeper constraint confirm delete dialog previously did not care whether you clicked cancel or delete, and deleted the constraint either way. Now only deletes it if you clicked Delete.
- Gatekeeper installation wizard copy to clipboard button is now responsive
- Kubesec report header now looks nicer on smaller screens
  - use app-ellipsis-text to restrict size of image name, adjusted spacing between text & score
  - accordion uses app-ellipsis-text as well (non-clickable). On smaller screens, the 'Score: X' will be displayed under the image name instead of next to it.
- gatekeeper constraint template editting. The code mirror now scrolls instead of wrapping for lines that are too long
- Image scan results issues table
  - scanner name, issue Title, and fixable column now employ app-ellipsis-text.
  - min & max widths adjusted on some columns
- historical vulnerabilities report table now has min width set on all columns
- vulnerability difference by date report page
  - namespace dropdown no longer half width on small screens
  - use app-ellipsis-component for namespaces & image columns
  - Use checkboxes instead of true/false for running and fixable columns
  - Set min widths for columns
# Misc
Corrected .main-container to have flex-direction rather than flex-display. (flex-display is not a valid CSS property. No regression as it was already getting flex-direction: row from elsewhere)


jacobbeasley (approved):
Nice and thank you for the thorough explanation!

PR #285, opened 2023-09-13
Determine Authority Levels based on AD groups (Ticket 6743)

kyle-intelletive:
# Extra fields for Azure OAuth config
Azure OAuth now has 2 extra (optional) fields, and they are configurable in the add/edit sign on method dialog.
## defaultAuthorityId
the ID of the authority level given to users not assigned an authority level by an AD group. Will be treated as if it was read only if not set.
## groupAuthorities 
An array of objects containing an authorityId and a groupId.
If a user is part of an AD group with an ID specified in this array, they are eligible for the associated authority.

# Granting Authorities
When a user signs in for the first time with Azure Auth, they will be created with the appropriate Authority Level based on these mappings. When a used signs in with their already created Azure account, their authority will be checked. If it is different than what it should be, it will be changed.

Either way the rules for determining which authority level is:
1. If the user is part of at least one AD group that is associated with an authority level, they will get the highest authority level amongst them. (Note that this can mean they have a lower privilege the defaultAuthorityId if they are part of only AD groups with lower authority levels)
2. If the user is not part of an AD group that grants privileges, and defaultAuthorityId is set on the config, they will receive that authority level.
3. If defaultAuthorityId is not set, and they aren't granted an authority from their AD groups, they will be Read Only users.

# Other
## Log in page saves auth method selection
Did for convenience in testing and thought it would be a feature. After changing the authentication method, will now save your selection to local storage and auto select it when you return to the sign in page.

## Use rows/cols for sign in method add/edit form
Converted the add/edit sign in method modal to use rows/cols instead of paragraphs, so that I could better make the styles for the additions of the new fields

### DIFF
--- a/dash/backend/src/modules/auth/services/oauth2/azure-oauth2.service.ts
+++ b/dash/backend/src/modules/auth/services/oauth2/azure-oauth2.service.ts
@@ -4,7 +4,7 @@ import { AxiosRequestConfig } from 'axios';
 import * as qs from 'qs';
 import { Oauth2AuthProvider } from './oauth2-auth-provider';
 import { ConfigService } from '@nestjs/config';
-import { OAuth2AuthStrategyConfig } from '../../models/auth-configuration';
+import { AuthConfiguration, AzureAuthStrategyConfig, OAuth2AuthStrategyConfig } from '../../models/auth-configuration';
 import { SourceSystem, UserAuthority, UserProfileDto } from '../../../user/dto/user-profile-dto';
 import {lastValueFrom} from 'rxjs';
 import {AuthorityId} from '../../../user/enum/authority-id';
@@ -13,6 +13,7 @@ import {DirectoryGroup} from '../interfaces/Azure/directory-group';
 
 @Injectable({scope: Scope.REQUEST})
 export class AzureOauth2Service extends Oauth2AuthProvider {
+  _authConfiguration: AuthConfiguration<AzureAuthStrategyConfig>;
 
   constructor(private readonly httpClient: HttpService,
               protected readonly configService: ConfigService,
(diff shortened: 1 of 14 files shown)


dash/backend/src/modules/auth/services/oauth2/azure-oauth2.service.ts
```diff
@@ -29,17 +31,25 @@ export class AzureOauth2Service extends Oauth2AuthProvider {
...
+
       if (users && Array.isArray(users) && users.length > 0) {
         if (!users[0].isActive) {
           throw new Error('This user is not active.');
         } else {
-          return users[0];
+          const user = users[0];
+          if (user.authorities?.length !== 1 || user.authorities[0].id !== expectedAuthority) {
+            user.authorities = [{ id: expectedAuthority }];
+            return this.userProfileService.updateUser(user).then(resp => resp[0]);
```
becky-intelletive, line 50: Should this be awaiting the promise's resolution?

dash/frontend/src/app/modules/public/pages/login/login.component.ts
```diff
@@ -44,18 +44,26 @@ export class LoginComponent implements OnInit {
...
-    }, () => {
-    });
+    this.authService.getAvailableAuthenticationMethods()
+      .pipe(take(1))
+      .subscribe({
+        next: (response: IServerResponse<IAuthenticationMethod[]>) => {
+          const authenticationMethods = response.data;
+          this.inSiteCredentialAuthenticationMethods = authenticationMethods;
+          const savedLoginId = this.retrieveLastAuthMethod();
+          const signInMethod = this.inSiteCredentialAuthenticationMethods.find(iscam => String(iscam.id) === savedLoginId)
```
becky-intelletive, line 54: iscam looks like iScam. like the scam version of an iPhone haha

becky-intelletive (approved):

dash/backend/src/modules/auth/services/oauth2/azure-oauth2.service.ts
kyle-intelletive: In this particular case, it does not matter.

As it currently is, it will return the promise, which will then be evaluated when this function is awaited.

If we added an await, it would evaluated that promise now, then wrap the result in a new promise that would be evaluated when this function is returned.


The times it would be different is if there was synchronous error handling happening (which in this case there is not).

For example, if X(), throws an error, that error will be caught by the try/catch in example A, but not in B as X will not actually throw its error until this function is awaited elsewhere.

```
// A
try {
  return await X();
} catch(err) {
   ...
}

// B
try {
  return X();
} catch (err) {
   ...
}



```

becky-intelletive (approved):

PR #289, opened 2023-09-19
Speed up "Vulnerability Difference by Date" Endpoints

becky-intelletive:
Frontend:
- report page: autofill dates for convenience

Backend:
- `image.dao.ts`: + subquery filtering to speed up graph
- `reports.dao.ts`: rework `getDifferencesInVulnerabilities` method
- `reports.service.ts`: spacing for `getVulnerabilityDifferenceByDate` method

### DIFF
--- a/dash/backend/src/modules/reports/dao/reports.dao.ts
+++ b/dash/backend/src/modules/reports/dao/reports.dao.ts
@@ -43,7 +43,7 @@ export class ReportsDao {
             .andWhere('i.deleted_at', null)
             .andWhere('c.deleted_at', null);
         if (clusterId) {
-            subQuery = subQuery.andWhere('i.cluster_id', clusterId);
+            subQuery = subQuery.andWhere('hki.cluster_id', clusterId);
         }
         if (namespaces) {
             subQuery = subQuery.whereIn('ki.namespace', namespaces);
@@ -552,7 +552,7 @@ export class ReportsDao {
         .whereIn('hki.saved_date', [startDate, endDate]);
 
       if (clusterId) {
-        imageDetailsQuery.andWhere('i.cluster_id', clusterId);
+        imageDetailsQuery.andWhere('hki.cluster_id', clusterId);
       }
       if (namespaces) {
         imageDetailsQuery.whereIn('hki.namespace', namespaces);
--- a/dash/backend/src/modules/reports/dao/reports.dao.ts
+++ b/dash/backend/src/modules/reports/dao/reports.dao.ts
@@ -531,13 +531,13 @@ export class ReportsDao {
         });
     }
 
-    async getVulnerabilityDifferenceSummaryByDate(
+    async getDifferencesInVulnerabilities(
       startDate: string, endDate: string, clusterId?: number, namespaces?: Array<string>,
       severityLevels?: Array<VulnerabilitySeverity>, fixAvailable?: string,
-    ): Promise<{ countOfFixedVulnerabilities: number, countOfNewVulnerabilities: number }> {
+    ): Promise<ReportsDifferenceByDateDto> {
       const knex = await this.databaseService.getConnection();
 
-      const image_details_query = knex.select([
+      const imageDetailsQuery = knex.select([
         'i.id',
         knex.raw("concat(i.name,':', i.tag) as image"),
         'i.running_in_cluster',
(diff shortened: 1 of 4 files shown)


dash/backend/src/modules/reports/dao/reports.dao.ts
```diff
@@ -531,113 +531,94 @@ export class ReportsDao {
...
+      const endDateAsDate = parseISO(endDate);
+      const validatedEndDate = endDateAsDate.toISOString().split('T')[0];
+      const startDateAsDate = parseISO(startDate);
+      const validatedStartDate = startDateAsDate.toISOString().split('T')[0];
+
+      const listsQuery = knex
+        .select(['scanner_name', 'image_id', 'image', 'running_in_cluster', 'type', 'severity', 'is_fixable', 'namespaces'])
+        .from(scanResultsQuery.as('scan_results'))
+        .whereRaw(knex.raw(`NOT saved_dates @> '{"${validatedStartDate}", "${validatedEndDate}"}'`));
+      const newVulnerabilities= listsQuery.clone().whereRaw(knex.raw(`NOT saved_dates @> '{"${validatedStartDate}"}'`));
```
kyle-intelletive, line 597: Starting here, you interpolate several values into knex.raw()'s.

I see that in this case those values are being generated so there shouldn't be a risk of SQL injection. It would still be a good idea to either add a comment making that clear, or to parameterize the knex.raw()'s

Parameterization example:
query.where(knex.raw('my_col = ? OR other_col = ?', [ val1, val2]));

dash/backend/src/modules/reports/dao/reports.dao.ts
```diff
@@ -531,113 +531,94 @@ export class ReportsDao {
...
+      ])
+        .from(scanResultsQuery.as('scan_results'));
+
+      return Promise.all([
+        newVulnerabilities.then(response => {
+          return plainToClass(ReportsVulnerabilityExportDto, response);
+        }),
+        fixedVulnerabilities.then(response => { return plainToClass(ReportsVulnerabilityExportDto, response); }),
+        summaryQuery,
+      ]).then((values) => {
```
kyle-intelletive, line 612: There is a typescript feature you can use here. I don't know what the proper name is, but if you have

(values: [var1, var2, ...]) => {...} as your callback definition, you can reference the elements passed in as the values array by the name. This may make this block easier to read if you use that to give the value meaningful names instead of accessing indices of the array.

dash/frontend/src/app/modules/private/pages/reports/vulnerability-difference-by-date/vulnerability-difference-by-date.component.ts
```diff
@@ -116,8 +127,8 @@ export class VulnerabilityDifferenceByDateComponent implements OnInit, OnDestroy
       namespaces: [[]],
       severityLevels: [[]],
       fixAvailable: [],
-      startDate: ['', [this.customValidatorService.dateInPastOrToday]],
-      endDate: ['', [this.customValidatorService.dateInPastOrToday]]
+      startDate: [this.twoDaysAgo, [Validators.required, this.customValidatorService.dateInPastOrToday]],
+      endDate: [this.yesterday, [Validators.required, this.customValidatorService.dateInPastOrToday]]
```
kyle-intelletive, line 131: We should probably have backend enforcement/validation or defaulting of these dates as well, as it looks like the query will bomb without them.

dash/backend/src/modules/reports/dao/reports.dao.ts
becky-intelletive: In this case, parameterization does not work. I tried a bunch of times haha. The quotes have to be exactly right and parameterization makes them unhappy.

becky-intelletive: Added comments.

dash/backend/src/modules/reports/dao/reports.dao.ts
becky-intelletive: If you remember how exactly to do that I'll gladly put it in! :)

In the meantime, I added declarations as the first part of the .then

dash/frontend/src/app/modules/private/pages/reports/vulnerability-difference-by-date/vulnerability-difference-by-date.component.ts
becky-intelletive: Yeah it will 😅

dash/backend/src/modules/reports/dao/reports.dao.ts
kyle-intelletive: Sorry for the confusion

kyle-intelletive (approved):

dash/backend/src/modules/reports/dao/reports.dao.ts
becky-intelletive: Darn! I was hoping there was a way to do it haha

PR #290, opened 2023-09-20
Prettify the Image Scan Results Print Page

becky-intelletive:
- hid certain buttons on print
- removed a full column on print (its button is hidden)

### DIFF
--- a/dash/frontend/src/app/modules/private/pages/image/image-scan-result/image-scan-result.component.ts
+++ b/dash/frontend/src/app/modules/private/pages/image/image-scan-result/image-scan-result.component.ts
@@ -351,6 +351,10 @@ export class ImageScanResultComponent implements OnInit, AfterViewInit, OnDestro
       });
   }
 
+  downloadReport() {
+    print();
+  }
+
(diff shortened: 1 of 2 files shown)


dash/frontend/src/app/modules/private/pages/image/image-scan-result/image-scan-result.component.ts
```diff
@@ -351,6 +351,10 @@ export class ImageScanResultComponent implements OnInit, AfterViewInit, OnDestro
       });
   }
 
+  downloadReport() {
```
kyle-intelletive, line 354: Is there a reason to have this function over calling print directly?

kyle-intelletive (approved):

dash/frontend/src/app/modules/private/pages/image/image-scan-result/image-scan-result.component.ts
becky-intelletive: Yeah, Angular doesn't believe in straight prints lol
image

PR #292, opened 2023-09-21
+ Security for Metrics Endpoint

becky-intelletive:
- Env vars added:
```
METRICS_DISABLE_ENDPOINT  # default: false
METRICS_SECURE_ENDPOINT  # default: false
METRICS_API_KEY  # default: false
METRICS_MINIMUM_AUTHORITY_LEVEL  # default: false
```
- Add API_KEY authority (including db migration)
- Add metrics to config service
- Add metrics guard
   - throw 404 if endpoint is disabled
   - if endpoint is secured,
      - if there's a logged-in user, check if they have the minimum required authority
      - if not, try to retrieve the api key and use the key to retrieve the user
         - if there's a user, use that to check if they have the minimum required authority
         - if there's no user, check if the api key is the same as the configured allowed key
      - checking the minimum required authority checks if the user has the authority defined in METRICS_MINIMUM_AUTHORITY_LEVEL, a greater authority, or the API_KEY authority
   - if endpoint is NOT secured, allow access
- `api-key.dao.ts`: add dao function to check if the api key exists
- Command module:
   - Add metrics key to the helm setup command
   - Add method & command to create an api key
- `LoggedInUserFactory.ts`: fix spacing (no other changes)

### DIFF
--- a/dash/backend/src/config/metrics.ts
+++ b/dash/backend/src/config/metrics.ts
@@ -1,0 +1,4 @@
+export default {
+  secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  placeholder: 'helloworld',
+};
--- a/dash/backend/src/config/metrics.ts
+++ b/dash/backend/src/config/metrics.ts
@@ -2,4 +2,8 @@ export default {
   disableEndpoint: (process.env.METRICS_DISABLE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
   secureEndpoint: true,
   // secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  security: {
+    apiKey: process.env.METRICS_API_KEY,
+    minimumAuthority: process.env.METRICS_MINIMUM_AUTHORITY_LEVEL,
+  },
 };
--- a/dash/backend/src/config/metrics.ts
+++ b/dash/backend/src/config/metrics.ts
@@ -1,6 +1,6 @@
 export default {
-  disableEndpoint: (process.env.METRICS_DISABLE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
-  secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  disableEndpoint: +process.env.METRICS_DISABLE_ENDPOINT || 0,
+  secureEndpoint: +process.env.METRICS_SECURE_ENDPOINT || 0,
   security: {
     apiKey: process.env.METRICS_API_KEY,
     minimumAuthority: process.env.METRICS_MINIMUM_AUTHORITY_LEVEL || 'SUPER_ADMIN',
--- a/dash/backend/src/config/metrics.ts
+++ b/dash/backend/src/config/metrics.ts
@@ -1,9 +1,8 @@
 export default {
   disableEndpoint: (process.env.METRICS_DISABLE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
-  secureEndpoint: true,
-  // secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
   security: {
     apiKey: process.env.METRICS_API_KEY,
-    minimumAuthority: process.env.METRICS_MINIMUM_AUTHORITY_LEVEL,
+    minimumAuthority: process.env.METRICS_MINIMUM_AUTHORITY_LEVEL || 'SUPER_ADMIN',
   },
 };
--- a/dash/backend/src/config/metrics.ts
+++ b/dash/backend/src/config/metrics.ts
@@ -1,4 +1,5 @@
 export default {
-  secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
-  placeholder: 'helloworld',
+  disableEndpoint: (process.env.METRICS_DISABLE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  secureEndpoint: true,
+  // secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
 };
--- a/dash/backend/src/modules/command-line/commands/helm-setup.command.ts
+++ b/dash/backend/src/modules/command-line/commands/helm-setup.command.ts
@@ -161,16 +161,15 @@ export class HelmSetupCommand {
       return true;
     }
 
+    // @TODO: clean up console statements when making cli commands silent
     async createServiceProfile(profileName: string, apiKey: string): Promise<boolean> {
       const titlecaseProfileName = profileName.toLowerCase().split(' ')
         .map(word => (word.charAt(0).toUpperCase() + word.slice(1)))
         .join(' ');
-      // const titlecaseProfileName = profileName.charAt(0).toUpperCase() + profileName.slice(1);
       const encryptedApiKey = await bcrypt.hash(apiKey, await bcrypt.genSalt(10));
 
       // check if api key already exists
       if (await this.apiKeyDao.apiKeyExists(apiKey)) {
-        // @TODO: clean up this message when making cli commands silent
         console.log('API Key is already in use.... skipping');
         return true;
       }
@@ -183,11 +182,9 @@ export class HelmSetupCommand {
       if (!userExists && profileName !== titlecaseProfileName) {
         user = await this.userDao.loadUser({email: titlecaseProfileName});
         userExists = !!user;
-        console.log('user 2: ', user);
       }
 
       if (userExists && user.length > 1) {
-        // @TODO: clean up this message when making cli commands silent
         console.log(`${titlecaseProfileName} has multiple associated users.... skipping`);
         return true;
       }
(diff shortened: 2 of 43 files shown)


dash/backend/src/config/metrics.ts
```diff
@@ -0,0 +1,8 @@
+export default {
+  disableEndpoint: (process.env.METRICS_DISABLE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
+  secureEndpoint: (process.env.METRICS_SECURE_ENDPOINT || 'false').toString().toLowerCase() === 'true',
```
kyle-intelletive, line 3: The rest of the config uses 1 or 0 to determine boolean config, this should be made consistent.

See database.migrations.enabled, email.system.enableErrorReportEmail, rabbitMQ.enabled for patterns (Annoyingly, they are all slightly different)

dash/backend/src/modules/command-line/commands/helm-setup.command.ts
```diff
@@ -14,11 +14,12 @@ import {generateRandomString} from "./generate-random-string";
 
 
 /**
- * These commands are used to setup initial m9sweeper data, such as the first admin user, when installing m9sweeper on
+ * These commands are used to set up initial m9sweeper data, such as the first admin user, when installing m9sweeper on
  * the kubernetes cluster
  */
 @Injectable()
 export class HelmSetupCommand {
+    promises: Promise<any>[] = [];
```
kyle-intelletive, line 22: It doesn't look you're actually using this new variable

dash/backend/src/modules/command-line/services/cli-command-builder.service.ts
```diff
@@ -61,6 +61,12 @@ export class CliCommandBuilderService {
                 'Accepts comma delimited list of namespaces to whitelist',
                 () => this.helmSetupCommand.loadDefaultNamespaceExceptions(),
             ),
+            new CliCommand(
```
kyle-intelletive, line 64: Why do we need a CLI command for this? We have a UI for creating API Keys.

dash/backend/src/modules/user/enum/Authority.ts
```diff
@@ -4,6 +4,7 @@ export enum Authority {
     READ_ONLY = 'READ_ONLY',
     KUBEHUNTER = 'KUBEHUNTER',
     KUBEBENCH = 'KUBEBENCH',
-    TRAWLER ='TRAWLER',
-    FALCO = 'FALCO'
+    TRAWLER = 'TRAWLER',
+    FALCO = 'FALCO',
+    API_KEY = 'API_KEY',
```
kyle-intelletive, line 9: This authority level should probably be called 'Metrics' or something similar.

It is very strange to have most API Keys not have the API Key level, and the Authority Level API KEY granting you access to the metrics endpoint is not intuitive.

kyle-intelletive (changes requested):
2 overall comments
1. It looks like you missed committing your migration
2. Is there a reason you are making these API Keys function so differently from the other ones?

dash/backend/src/modules/command-line/services/cli-command-builder.service.ts
becky-intelletive: Eh, fair enough. Though we could ask the same question of the runSeed() job.

dash/backend/src/modules/user/enum/Authority.ts
becky-intelletive: We don't really have anywhere else we use api keys. In my mind, this was a way to start granting them access to things.

The UI attaches API Keys to specific user profiles. This approach is more of a service-based API Key: giving other programs access to endpoints.

becky-intelletive:
@kyle-intelletive -
1. I didn't commit it because I didn't want to add to the migration tree yet.
2. See the comment in the guard. I think we could move that functionality into the logged-in user population file. It would allow us to pass in API keys as bearer tokens or headers. Beyond that, it's another programmatic API Key. Those are handled differently than users' individual API keys (which are attached to their permissions). I want to open access to the endpoint to different classes of "users":
   - ones with defined access levels who are logged in
   - ones with defined access levels who are not logged in but are using individual API keys (granted via the ui)
   - services that are using their assigned API key
   - services that are using the global API key

dash/backend/src/modules/user/enum/Authority.ts
becky-intelletive: I can change it so it's more specific. I just wanted to start making a path forward for programmatic access to endpoints.

becky-intelletive: Instead of having to make an endless number of Authorities based on specific criteria ;)

becky-intelletive:
@kyle-intelletive 
- Moved more validation from the guard to the LOGGED_IN_USER provider
   - LOGGED_IN_USER provider can now get api keys from bearer tokens and headers
- Authority.METRICS instead of Authority.API_KEY
- Checked in the migration
- Added a note to "save kubehunter report" to utilize the LOGGED_IN_USER provider

Any more thoughts on how it can be simplified/genericized?

kyle-intelletive:
> @kyle-intelletive
> 

>       
>       * LOGGED_IN_USER provider can now get api keys from bearer tokens and headers
> 
>     * Authority.METRICS instead of Authority.API_KEY
> 
>     * Checked in the migration
> 
>     * Added a note to "save kubehunter report" to utilize the LOGGED_IN_USER provider
> 
> 
> Any more thoughts on how it can be simplified/genericized?
Given that the LOGGED_IN_USER is factory, it will be kind of difficult 

I can think of a couple ways. Here are some quick overviews of them, I can discuss them in more detail at some point if you'd like.

# A. Change to service
If we use a service rather than factory (still scoped to request) - we could implement a decorator like below that would basically just call that service with the appropriate config. then we'd change our references to call this.LOGGED_IN_USER.get() rather than it being a defined variable. We'd also have to ensure that the Authentication() decorator is resolved before any other decorators that care about the user 

# B. Make mono-decorator
Basically, we roll the functionality of UseGaurds, AllowedAuthorityLevels, and LOGGED_IN_USER into one decorator.
then, we'd also have to define another decorator that functions like the @Body() decorator, that pulls the value of the user from the metadata that the main decorator will call. (Won't need that secondary decorator if we only care that the user is authenticated and not who they are)

# C. Provide variations of the LOGGED_IN_USER
We can inject variations of the LOGGED_IN_USER factory that handle auth differently. Off the top of my head, I know we can do this at a module level, we'd have to do more research to see if it can be changed at the class/function level 


# Sample code for both A & B
```
@Authentication({
 authorityLevels: [Authority.Admin ...]
  // If true, checks for bearer token as "normal user"
  allowUsers: true, 
  // If true, checks for API Key users
  allowApiKeys: true, 
  // optional, checks some reasonable default if not set
  apiKeySources: {
     // Checks in the headers and query params of provided names
     headers: ['x-api-key', 'whatever-else' .....]
     queryParams: ['key', 'apiKey' ....],
     // Check if bearer token is actually an api key
     bearerToken: true
  }
})
myEndpoint() {
...
}
```

becky-intelletive:
> > @kyle-intelletive
> I can think of a couple ways. Here are some quick overviews of them, I can discuss them in more detail at some point if you'd like.
> ...
That seems like an expansion of scope to improved the way that the app fills/handles LOGGED_IN_USER (and subsequent authorization). Do you think this would make more sense as a new ticket?

kyle-intelletive:
> > > @kyle-intelletive
> > > I can think of a couple ways. Here are some quick overviews of them, I can discuss them in more detail at some point if you'd like.
> > > ...
> > > That seems like an expansion of scope to improved the way that the app fills/handles LOGGED_IN_USER (and subsequent authorization). Do you think this would make more sense as a new ticket?

@becky-intelletive 
For sure, this would definitely be a different feature. I will review the changes made from the other topics shortly.

becky-intelletive:
> @becky-intelletive For sure, this would definitely be a different feature. I will review the changes made from the other topics shortly.

https://dev.azure.com/intelletive/Minesweeper/_boards/board/t/Minesweeper%20Team/Issues/?workitem=6897

kyle-intelletive (dismissed):

dash/backend/src/modules/kube-hunter/dto/kube-hunter-raw.interface.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+interface Options {
+  enableCircularCheck: boolean
+  enableImplicitConversion: boolean
+  excludeExtraneousValues: boolean
+  exposeDefaultValues: boolean
+  exposeUnsetFields: boolean
+  ignoreDecorators: boolean
+}
```
kyle-intelletive, line 80: A lot of the interfaces in this file are very generically named. It would probably be a good idea to name them something like KubeHunterWhatever.

Right now all the generic ones are just used in IKubeHunterRawResponseSection, but in the future, if we have a reason to do more processing on the reports, we may need to export these inner interface

manifests/charts/dash/templates/prometheus-rules-default.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       rules:
       - alert: ManyNewCVEs
         annotations:
-          description: Fires when the number of new major + critical CVEs in the last 10m exceeds 20
+          description: Fires when the number of new major + critical CVEs in the last 10m exceeds {{ .threshold }}
```
kyle-intelletive, line 20: Should this have use a .duration like the ones below?

becky-intelletive: The .duration field is actually new

dash/backend/src/modules/kube-hunter/dto/kube-hunter-raw.interface.ts
becky-intelletive: So I generated the interfaces by throwing the response into an interface generator ;)
I had made the names slightly more explicit and consolidated repeat interfaces, but I hadn't bothered making them too clear since they were internal interfaces.
The latest commit adds `IKHRawResponse` to the beginning of each interface.

becky-intelletive:
@kyle-intelletive - ready for re-review

kyle-intelletive (approved):

PR #296, opened 2023-09-26
Auth audit

kyle-intelletive:
# Changes made from auth audit
- Gatekeeper (All cahnges for RO users)
   - Get Constraint Template endpoint made accessible to RO users
   - Constraint Details page - Edit button becomes 'View' button for viewing the YML of the template, and opened dialog 
is read only
   - Gatekeeper not installed - no longer has install button in recommendation section
   - Gate keeper installed with no constraint templates - no longer has add button in recommendations section
- Reports - All endpoints made accessible to RO users
- 'Add Cluster Group' button in side nav no longer visible for RO users
- kube-bench - no longer has 'run audit button' for RO users
- kube-hunter - no longer has 'run audit button' for RO users
- Exceptions Details - edit & delete button no longer visible for RO users
- exceptions list - action column no longer visible to RO users
- Falco - Install button no longer visible for RO users
- Image scan results - (re)-scan button no longer visible for RO users
- policy list - actiosn no longer visible for RO users

# Misc changes made
- Gave color to submit button on reset password form
- Gatekeeper constraint template details - After saving edits to the constraint template, the table of constraints would get cleared. Now that does not happen
- Fixed race condition in AlertDialogComponent that could cause the page to navigate before the web request completed. This result in behaviors such as deleting a cluster, and the redirect to the cluster group page would happen before the cluster was deleted - resulting in the deleted cluster still being shown on that page.

### DIFF
--- a/dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.html
+++ b/dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.html
@@ -39,11 +39,11 @@
 
               <div class="col-xs-12 col-sm-12 col-md-2 col-lg-2">
                 <div class="text-align-right">
-                  <button  mat-raised-button color="primary" (click)="openInstallWizard()" *ngIf="gatekeeperConstraintTemplates === null; else recommendationButton">
+                  <button mat-raised-button color="primary" [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" (click)="openInstallWizard()" *ngIf="gatekeeperConstraintTemplates === null; else recommendationButton">
                     Install
                   </button>
                   <ng-template #recommendationButton>
-                    <button  mat-raised-button color="primary" (click)="openAddConstraintDialog()" *ngIf="gatekeeperConstraintTemplates?.length === 0">
+                    <button  mat-raised-button color="primary" [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" (click)="openAddConstraintDialog()" *ngIf="gatekeeperConstraintTemplates?.length === 0">
                       Setup
                     </button>
                   </ng-template>
--- a/dash/frontend/src/app/modules/private/pages/exceptions/exception-list/exception-list.component.html
+++ b/dash/frontend/src/app/modules/private/pages/exceptions/exception-list/exception-list.component.html
@@ -41,9 +41,9 @@
             </mat-cell>
           </ng-container>
 
-          <ng-container matColumnDef="edit" [allowedRoles]="['ADMIN', 'SUPER_ADMIN']">
-            <mat-header-cell *matHeaderCellDef>Actions</mat-header-cell>
-            <mat-cell *matCellDef="let exception">
+          <ng-container matColumnDef="edit">
+            <mat-header-cell [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" *matHeaderCellDef>Actions</mat-header-cell>
+            <mat-cell [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" *matCellDef="let exception">
               <a mat-icon-button [routerLink]="['/private', 'exceptions', exception?.id, 'edit']">
                 <mat-icon>create</mat-icon>
               </a>
--- a/dash/frontend/src/app/modules/shared/alert-dialog/alert-dialog.component.ts
+++ b/dash/frontend/src/app/modules/shared/alert-dialog/alert-dialog.component.ts
@@ -30,14 +30,15 @@ export class AlertDialogComponent {
             if (!environment.production) {
               console.log(result);
             }
+
+            this.matDialogRef.close(true);
+            if (this.data?.afterRoute?.length > 0) {
+              this.router.navigate(this.data.afterRoute);
+            } else if (this.data?.reload) {
+              window.location.reload();
+            }
           }
         });
-    }
-    this.matDialogRef.close(true);
-    if (this.data?.afterRoute?.length > 0) {
-      this.router.navigate(this.data.afterRoute);
-    } else if (this.data?.reload) {
-      window.location.reload();
     }
   }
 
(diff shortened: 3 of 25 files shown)


dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.html
```diff
@@ -39,11 +39,11 @@ <h2 class="text-style-green" *ngIf="gatekeeperStatusLoaded && gatekeeperInstalla
 
               <div class="col-xs-12 col-sm-12 col-md-2 col-lg-2">
                 <div class="text-align-right">
-                  <button  mat-raised-button color="primary" (click)="openInstallWizard()" *ngIf="gatekeeperConstraintTemplates === null; else recommendationButton">
+                  <button mat-raised-button color="primary" [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" (click)="openInstallWizard()" *ngIf="gatekeeperConstraintTemplates === null; else recommendationButton">
```
becky-intelletive, line 42: Were you able to test that this doesn't show up when for READ_ONLY users?

dash/frontend/src/app/modules/shared/alert-dialog/alert-dialog.component.ts
```diff
@@ -30,15 +30,16 @@ export class AlertDialogComponent {
             if (!environment.production) {
               console.log(result);
             }
+
```
becky-intelletive, line 33: If there's no function to run, will this still close?
atm it looks like if there's no function, it won't close at all.

kyle-intelletive: You may be right, I will verify tomorrow

dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.html
kyle-intelletive: Yes,
unsure if your question is more aimed at the allowedRoles or the gatekeeper setup.

I ran against my local docker desktop cluster, so I was able to clear my gatekeeper and check this page before & after installing.

The allowedRoles decorator basically adds 'display: none' to its parent element when the currently signed in user does not have one of the specified roles

dash/frontend/src/app/modules/shared/alert-dialog/alert-dialog.component.ts
kyle-intelletive: The modal always closed, because the submit button had the mat-dialog-close property. In the latest commit, I made it so it will run this section either after the observable to run, or if there is no observeable.

I checked all the spots that it is used and it works as I would expect in each.

I think that this alert dialog should probably be made to only have 1 "mode" as a tech debt item, as right now, it can be running an observeable or not, and can navigate to a page, refresh the page, or do neither.

It probably should just return the boolean and let the calling page handle making the observable calls and any navigation that needs to happen as a result afterwards.

dash/frontend/src/app/modules/private/pages/cluster/gate-keeper/gate-keeper.component.html
becky-intelletive: The allowedRoles :)
Did you sign in as the READ_ONLY user and make sure? I had previously noticed that the decorator wasn't always 100% effective.

kyle-intelletive: Yes, the allowed Roles always works as long as it is in the correct element. I think I know exactly what circumstance you are thinknig it doesn't work in, because I fixed a spot where it looked like it wasn't working, and Git said you added it. I'll comment on that spot and explain.

dash/frontend/src/app/modules/private/pages/exceptions/exception-list/exception-list.component.html
```diff
@@ -41,9 +41,9 @@
             </mat-cell>
           </ng-container>
 
-          <ng-container matColumnDef="edit" [allowedRoles]="['ADMIN', 'SUPER_ADMIN']">
-            <mat-header-cell *matHeaderCellDef>Actions</mat-header-cell>
-            <mat-cell *matCellDef="let exception">
+          <ng-container matColumnDef="edit">
+            <mat-header-cell [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" *matHeaderCellDef>Actions</mat-header-cell>
+            <mat-cell [allowedRoles]="['ADMIN', 'SUPER_ADMIN']" *matCellDef="let exception">
```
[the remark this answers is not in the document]
becky-intelletive: In this case, would it make more sense to change the list of displayed columns in the ts file?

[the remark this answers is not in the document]
becky-intelletive: (There were also several buttons and such that weren't configured properly. I switched my user role to make absolutely certain that the allowedRoles was working and it wasn't.)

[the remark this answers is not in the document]
kyle-intelletive: Do you have an example of where it wasn't working? I've tested everything I changed and they work as expected.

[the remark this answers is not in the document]
kyle-intelletive: And doing this or changing the displayedColumns would both get to the same point, this was just a little quicker & easier to get done. I can change it if you think that would be better

becky-intelletive (approved):

dash/frontend/src/app/modules/private/pages/exceptions/exception-list/exception-list.component.html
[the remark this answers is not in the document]
becky-intelletive: I also personally prefer an *ngIf over allowedRoles - I don't like to leave buttons in the DOM if they shouldn't be there.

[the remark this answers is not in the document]
becky-intelletive: m9sweeper is for developers, who likely would have a background in DOM manipulation :)

PR #313, opened 2023-11-07
Security Audit Pdf Report

kyle-intelletive:
Ticket: https://dev.azure.com/intelletive/Minesweeper/_boards/board/t/Minesweeper%20Team/Issues/?workitem=6441

Allows generating security audit reports.

# UI
- Under reports added Security Audit Report
- Can select clusters, namespaces, and tools to include
  - By default, includes current cluster & all namespaces/tools 
- Generating report will cause report to run & download in browser

# Under the hood
- Using pdfmake npm package
- Added security audit report module
- High level flow
   - audit report service retrieves objects to run report on.
   - For each cluster, it is passed to the cluster service
   - That cluster is run through the service for each enabled tool
   - The wrapper function assembles everything into the pdf
     - Intro section w/ Table of Contents
     - Overview with a quick summary for each tool with
     - (very) detailed section for each cluster that goes into extreme details for each of the enabled tools
- The tool services have a simple interface that should make it fairly easy to add a new tool. Steps to do so are:
  - Add entry to both frontend & backend tool enum 
  - Implement tool service implementing the interface
  - Add tool to the toolService factory's getTool function.


dash/backend/src/modules/kubesec/services/kubesec.service.ts
```diff
@@ -36,17 +36,13 @@ export class KubesecService {
         return podList;
     }
 
-    async runKubesecByPod(podName: string, namespace: string, clusterId: number): Promise<any> {
+    async runKubesecByPod(podName: string, namespace: string, clusterId: number): Promise<{ data: KubesecScanResults }> {
         const cluster = await this.clusterService.getClusterById(clusterId);
         if (cluster) {
```
kyle-intelletive, line 41: TLDR: removing the try-catch is not a regression.

The try-catch here effectively caused this to return the string "Pod couldn't be reached". The only previous usage of this function was the kube sec controller, which wouold then immediately error trying to access a property of that string. So removing the try-catch causes it to catch the real error at that point.

becky-intelletive (dismissed):

dash/frontend/src/app/modules/private/pages/reports/security-audit-report/security-audit-report.component.html
```diff
@@ -0,0 +1,75 @@
...
+            </form>
+          </div>
+        </mat-card-content>
+      </mat-card>
+    </div>
+  </div>
+</div>
+
+
+<p>printable-audit-report works!</p>
```
jshoberg, line 74: Is this line meant to be in the PR?

kyle-intelletive: Nice catch, removed in latest commit

becky-intelletive (dismissed):

jshoberg (approved):

PR #315, opened 2023-11-08
now resets pod links when syncing

kyle-intelletive:
Ticket: https://dev.azure.com/intelletive/Minesweeper/_boards/board/t/Minesweeper%20Team/Issues/?workitem=6925

Previously calculated which images were new to the pod, and added those links.

Now as a transaction clears all links, and then recreates with the current data.

### DIFF
--- a/dash/backend/src/modules/pod/dao/pod.dao.ts
+++ b/dash/backend/src/modules/pod/dao/pod.dao.ts
@@ -339,21 +339,24 @@ export class PodDao {
     async linkExistingPodImages(podId: number, imageIds: Array<number>): Promise<void> {
       const knex = await this.databaseService.getConnection();
 
-      const links = imageIds.map(id => ([podId, id]));
+      const links = imageIds.map(id => ({pod_id: podId, image_id: id}));
 
-      let existingLinks = await knex('pod_images').whereIn(['pod_id', 'image_id'], links) as [{pod_id: number, image_id: number}];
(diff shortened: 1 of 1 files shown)


dash/backend/src/modules/pod/dao/pod.dao.ts
```diff
@@ -339,21 +339,24 @@ export class PodDao {
...
-      const links = imageIds.map(id => ([podId, id]));
-
-      let existingLinks = await knex('pod_images').whereIn(['pod_id', 'image_id'], links) as [{pod_id: number, image_id: number}];
+      const links = imageIds.map(id => ({pod_id: podId, image_id: id}));
 
-      let newLinks: {pod_id: number,  image_id: number}[] = [];
+      await knex.transaction(async function(trx){
+        try {
+          // Clear all existing links to the pod
+          await knex.delete()
```
becky-intelletive, line 347: Is there a history table these should go into or no?

kyle-intelletive: This is my understanding, let me know if this is incorrect or if you think we should confirm with Jake:
The sync job this is part of runs (by default) every 30minutes, and makes syncs all of the kubernetes data that is not part of the history.

The history job runs (by default) daily at 12:01, and saves as a point in time what is currently exists fromt he sync jobs.

becky-intelletive (approved):

Issue #303, opened 2023-10-18
trawler unable to scan private aws ecr images


sumagoudb:
I have added docker registry for my ecr with aws iam secret id and token, with auth type Amazon Container Registry.
After that m9sweeper stopped sending scan jobs to `trawler`, in trawler logs i can see it is waiting for the scan job.

I have also tried enabling debug logs but no luck. I do not see any errors too.

Please help.

jacobbeasley:
Without more context its hard to comment on precisely what is going on in your case, but I'll explain how this feature works and that may help you figure out your issue. 

**How Scans are Queued:** 
1. Every hour, a scrape job runs to scrape and store a record of all namespaces, pods, and images running in your cluster. It will queue up scans for any unscanned image. 
2. Every night, a scrape job runs to store a historical record of all namespaces, pods, and images running in your cluster. If any images have scans but the rescan period configured in your cluster configuration/scanner configuration has been reached, it will queue up images to be rescanned
3. You can manually add images by clicking in the add button on the images page (the + button at the top) or trigger them to be rescanned in the UI by navigating to an image and then clicking the rescan button at the top. 

**How scans run:** 
1. dash writes to rabbitmq topic
2. trawler reads from rabbitmq topic and runs the scan

Based upon your logs just showing "waiting for scan job" I'm wondering if maybe dash isn't writing images to be scanned, or maybe the next rescan period hasn't been reached yet. I recommend trying to manually rescan an image by clicking the rescan button in the UI.

sumagoudb:
unfortunately i cannot login anymore because of issue https://github.com/m9sweeper/m9sweeper/issues/305, may be I have to redeploy.

here is the behaviour:

1. Right now we added ACR using helm chart values during the installation.
2. We can see all the images scanned except Amazon private ECR images.
3. when we click on a image we see loading top right corner, the scan image option greyed out. Not able to stop the scan or restart scan as it is greyed out.
4. no exceptions or errors in dash or trawler.

sumagoudb:
I am now able to login and when I click on the image to be scanned, I see the wheel spinning. its not able to scan any image.

PR #321, opened 2023-12-01
Add annotations to service account to allow adding eks role

sumagoudb:
Hello,

This changes should allow add annotations to the service account, which enables using AWS IAM role which can be assumed to scan images from AWS ECR.

I tried this workaround in our setup and it works. only catch is we need to have a dummy place holder for our ecr registry url in the m9sweeper docker registry list.

this should fix issue with https://github.com/m9sweeper/m9sweeper/issues/303


joffreychambrin:
I can confirm I had the same issue, and it worked well with your change

Lincon-Freitas:
What else do we need to merge this?

kyle-intelletive (approved):

kyle-intelletive:
This has been merged in, and will be a part of our next release.

Could you provide a link to the AWS documentation about the annotations you used so that we can add that link to our documentation?

PR #337, opened 2024-02-05
Compliance results map tests/cleanup

kyle-intelletive:


dash/backend/src/modules/policy/entities/compliance-result-map.ts
```diff
@@ -1,47 +1,73 @@
...
+/**
+ * Interface for a flattened issue from the map.
+ */
 export interface ComplianceIssue {
   policyId: number;
   scannerId: number;
   cve: string;
   compliant: boolean;
   severity?: string;
+  reason?: string;
```
kyle-intelletive, line 21: Note: this fields (and including it in getAllIssues() is actually a change. This does not affect any current places using this fuctionality.

It made sense to include since the reason being passed in for issues could be potentially useful in future improvements utilizing this function.

dash/backend/src/modules/policy/entities/compliance-result-map.ts
```diff
@@ -105,61 +152,51 @@ export class ComplianceResultMap {
     return true;
   }
 
-  /**
```
kyle-intelletive, line 108: This was to be with the other setResultForX functions

dash/backend/src/modules/policy/entities/compliance-result-map.ts
```diff
@@ -1,47 +1,73 @@
+/**
```
kyle-intelletive, line 1: The only functional change in this file was to add reason to ComplianceIssue and modify getAllIssues() to return that info.

The rest is comments/formatting

becky-intelletive (approved):

PR #342, opened 2024-03-04
get local running again; minor changes

becky-intelletive:
- ignore frontend vendor node_modules
- update `start:local` to use `nest start --watch` like it used to
   - bypasses the js compilation
   - allows proxy to work properly
- IDE's automatic whitespace removal
   - dash/backend/src/modules/command-line/services/kubernetes-api.service.ts
   - dash/backend/src/modules/security-audit-report/services/security-audit-kube-bench.service.ts
- "list k8s namespaces" call raises a 500 if there's an error (and frontend handles the error)
- "Kube-Bench" --> "kube-bench"
- add "Section " before section id in security audit report (I found it confusing)


manifests/charts/dash/templates/job-init.yaml
```diff
@@ -4,7 +4,7 @@ kind: Job
 metadata:
   name: {{ include "dash.fullname" . }}-init
   annotations:
-    "helm.sh/hook": post-install
+    "helm.sh/hook": post-install,post-upgrade
```
becky-intelletive, line 7: I had this locally and it's ringing a bell that we may have needed it. Any ideas why?
@jacobbeasley @kyle-intelletive

kyle-intelletive: With just post install, this job will only run after the initial helm install.

With both, it will run after the initial helm install and every helm upgrade.

kyle-intelletive (dismissed):

kyle-intelletive (approved):

